PR #37, opened 2021-08-24
fix/login-button-blocked

zhihaoSAR:


sg-gs (approved):

sg-gs:
@zhihaoSAR 
Changing `signing` function on `user.actions.ts` breaks SignUp functionality. 

For more info, see [Register component](https://github.com/internxt/drive-mobile/blob/0cafee0e9df478f14c6c7bb05ae7cca0da896b9e/src/screens/Register/index.tsx#L104)

PR #94, opened 2022-05-18
_: Refactor SortModal component to be more reusable, avoid redux inte…

PixoDev:
This PR refactors the SortModal component so it can be reused across the application, Redux internal component usage was removed totally so the component now can be controlled by the parent component and not by Redux

### DIFF
--- a/src/screens/DriveScreen/index.tsx
+++ b/src/screens/DriveScreen/index.tsx
@@ -22,7 +22,7 @@ import { ArrowDown, ArrowUp, CaretLeft, DotsThree, MagnifyingGlass, Rows, Square
 import { asyncStorage } from '../../services/asyncStorage';
 import { NativeStackNavigationProp } from '@react-navigation/native-stack';
 import { DriveListType, SortDirection, SortType } from '../../types/drive';
-import SortModal from '../../components/modals/SortModal';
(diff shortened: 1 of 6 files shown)


src/screens/DriveScreen/index.tsx
```diff
@@ -21,23 +21,35 @@ import AppScreen from '../../components/AppScreen';
...
+import { DriveListType, SortDirection, SortType } from '../../types/drive';
+import SortModal from '../../components/modals/SortModal';
+import fileService from '../../services/file';
 
 function DriveScreen(): JSX.Element {
   const navigation = useNavigation<NativeStackNavigationProp<any>>();
   const route = useRoute();
   const dispatch = useAppDispatch();
+  const [sortModalOpen, setSortModalOpen] = useState(false);
+  const [sortOptions, setSortOptions] = useState({
```
carlossalasamper, line 33: To put some downside, this variable should be called sortMode for consistency with the type it meets and we avoid the plural for an object as well

carlossalasamper (approved):
👍

src/screens/DriveScreen/index.tsx
PixoDev: yep, that makes more sense

PR #121, opened 2022-08-26
[MT-19]: Optimize filesystem storage usage from photos

PixoDev:
This PR provides:

- Better filesystem usage
- Improved photos stability and added upload time monitoring
- Refactorized tmp directory usage, now the tmp dir is deleted on app start so we free space used by the old session

### DIFF
--- a/src/components/gallery-views/GalleryAllView.tsx
+++ b/src/components/gallery-views/GalleryAllView.tsx
@@ -12,8 +12,10 @@ import { useTailwind } from 'tailwind-rn';
 
 const COLUMNS = 3;
 const GUTTER = 3;
-const GalleryAllView: React.FC<{ onLoadNextPage: () => Promise<void> }> = ({ onLoadNextPage }) => {
-  const tailwind = useTailwind();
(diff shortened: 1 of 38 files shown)


src/components/gallery-views/GalleryAllView.tsx
```diff
@@ -79,37 +81,34 @@ const GalleryAllView: React.FC<{ onLoadNextPage: () => Promise<void> }> = ({ onL
...
-      showsVerticalScrollIndicator
-      indicatorStyle={'black'}
-      refreshControl={
-        <RefreshControl
-          refreshing={refreshing}
-          onRefresh={async () => {
-            setRefreshing(true);
+  async function handleRefresh() {
+    setRefreshing(true);
+    await onRefresh();
```
sg-gs, line 86: What is going to happen if this method fails? Would the gallery keep loading infinitely instead of notifying the user?

sg-gs (approved):

src/components/gallery-views/GalleryAllView.tsx
PixoDev: this method is handled by the parent, the parent calls a redux thunk

In case the method fails, the UI will keep the data in memory, so the old photos rendered will appear

PR #131, opened 2022-09-19
[_]:(chore) New services structure based on products

PixoDev:
This PR provides a first iteration of a new folder structure for mobile services

**Motivation:**

We need some modularization in mobile, since it contains basically more than one Internxt products.

**Solution:**

Provide a new service folder organization based on products.

services
-- drive
-- photos
-- common

Each service file is contained in a folder, the folder ideally should have this structure in the future:

drive
-- file
---- driveFile.service.ts
---- driveFile.spec.ts
---- index.ts

Benefits:

- Modularized code, scoped by products
- Each file has their own test file, which ideally serve as documentation of the purpose of that file
- We can remove/add an entire product easily using this structure


masterprog-cmd (approved):

sg-gs:
I have to thank you for the comment on the PR, as is a very descriptive one which could be useful in case we have to backtrack any issue.

src/services/drive/folder/driveFolder.service.ts
```diff
@@ -68,7 +68,7 @@ class DriveFolderService {
         const fileFullName = `${file.name}${file.type ? '.' + file.type : ''}`;
         const relativePath = `${currentFolder.relativePath}/${fileFullName}`;
 
-        fileService.renameFileInNetwork(file.fileId, bucketId, relativePath);
+        driveFileService.renameFileInNetwork(file.fileId, bucketId, relativePath);
```
sg-gs, line 71: shouldn't this be - according to what has been done here - ? `drive.file.rename...`

src/services/drive/folder/driveFolder.service.ts
```diff
@@ -50,7 +50,7 @@ class DriveFolderService {
 
     while (pendingFolders.length > 0) {
       const currentFolder = pendingFolders[0];
-      const folderContentResponse = await fileService.getFolderContent(currentFolder.folderId);
+      const folderContentResponse = await driveFileService.getFolderContent(currentFolder.folderId);
```
sg-gs, line 53: here too: `drive.file.getFolderContent` ?

sg-gs (changes requested):

src/services/drive/folder/driveFolder.service.ts
PixoDev: Missed this ones, since they are relative, updating

sg-gs (approved):

PR #132, opened 2022-09-21
[MT-1]:(feature) New shared links compatible with new drive-web shared links

PixoDev:
This PR includes:

- An entire refactor of the Recents and Shared screen logic (they don't use Redux anymore, the state is now local and not global, which reduces the amount of extra renders)
- New shared links features (share the link, generate it, and delete it) using Internxt SDK
- Shared links for folders
- New UI implementations


src/screens/HomeScreen/index.tsx
```diff
@@ -26,12 +58,30 @@ const HomeScreen = (): JSX.Element => {
     {
       id: HomeTab.Recents,
       title: strings.screens.recents.title,
-      screen: <RecentsScreen searchText={searchText}></RecentsScreen>,
+      screen: (
+        <RecentsScreen
+          searchText={searchText}
+          isLoading={recentsLoading}
+          recentItems={recentItems}
+          refreshRecentItems={async () => {
```
sg-gs, line 66: `async-await` is not required here as `refreshRecentItems` is already `async`, therefore you can do:
`refreshRecentItems={refreshRecentItems} `

sg-gs: In case that does not fit with correct behavior, you can just use:
`refreshRecentItems={() => refreshRecentItems()}`

src/screens/HomeScreen/index.tsx
```diff
@@ -26,12 +58,30 @@ const HomeScreen = (): JSX.Element => {
...
       id: HomeTab.Shared,
       title: strings.screens.shared.title,
-      screen: <SharedScreen searchText={searchText}></SharedScreen>,
+      screen: (
+        <SharedScreen
+          searchText={searchText}
+          isLoading={sharedLoading}
+          sharedLinks={sharedLinks}
+          refreshSharedLinks={async () => {
+            await refreshSharedLinks();
```
sg-gs, line 81: same here related to the async-await usage topic

src/services/drive/share/driveShare.service.ts
```diff
@@ -1,47 +1,77 @@
...
-          type: NotificationType.Error,
-        });
-      }
+  public async shareGeneratedSharedLink(link: string) {
+    return Share.open({
+      message: link,
     });
+  }
+
+  public async deleteShareLink({ shareId }: { shareId: string }) {
```
sg-gs, line 52: `public` and `async` are not required here

src/services/drive/share/driveShare.service.ts
```diff
@@ -1,47 +1,77 @@
...
 
-        share.fileInfo.name = decryptedName;
-      } catch (err) {
-        const castedError = errorService.castError(err);
-        notificationsService.show({
-          text1: 'Error decrypting files: ' + castedError.message,
-          type: NotificationType.Error,
-        });
-      }
+  public async shareGeneratedSharedLink(link: string) {
```
sg-gs, line 46: `public` and `async` are not required here

src/useCases/common/shareFile.ts
```diff
@@ -0,0 +1,21 @@
+import { DisplayableError } from '@internxt-mobile/services/common';
+
+import fileSystemService from '@internxt-mobile/services/FileSystemService';
+import { FileSystemRef } from '@internxt-mobile/types/index';
+import strings from 'assets/lang/strings';
+
+export const shareFile = async ({ title, filePath }: { title: string; filePath: FileSystemRef }) => {
+  try {
+    return await fileSystemService.shareFile({
```
sg-gs, line 9: `await` is not required here

src/useCases/common/shareFile.ts
```diff
@@ -0,0 +1,21 @@
+import { DisplayableError } from '@internxt-mobile/services/common';
+
+import fileSystemService from '@internxt-mobile/services/FileSystemService';
+import { FileSystemRef } from '@internxt-mobile/types/index';
+import strings from 'assets/lang/strings';
+
+export const shareFile = async ({ title, filePath }: { title: string; filePath: FileSystemRef }) => {
```
sg-gs, line 7: `async` is not required here, I would say that is better to just define the `return` type of the function, even if it appears now obvious, as in the future someone could change it.

src/useCases/drive/loadRecentItems.ts
```diff
@@ -0,0 +1,16 @@
+import { DisplayableError } from '@internxt-mobile/services/common';
+import drive from '@internxt-mobile/services/drive';
+import strings from 'assets/lang/strings';
+
+export const loadRecentItems = async () => {
+  try {
+    return await drive.recents.getRecents();
```
sg-gs, line 7: `await` is not required here

src/useCases/drive/loadRecentItems.ts
```diff
@@ -0,0 +1,16 @@
+import { DisplayableError } from '@internxt-mobile/services/common';
+import drive from '@internxt-mobile/services/drive';
+import strings from 'assets/lang/strings';
+
+export const loadRecentItems = async () => {
```
sg-gs, line 5: `async` is not required here. Better define the `return` type of the function

sg-gs (approved):

PR #134, opened 2022-09-22
[_]:(chore) New services structure based on products for photos

PixoDev:
Same as https://github.com/internxt/drive-mobile/pull/131 but for photos


src/services/photos/utils/index.ts
```diff
@@ -0,0 +1 @@
+export * from './photosUtils';
```
sg-gs, line 1: if the utils files are inside the `photos/utils` folder, maybe is better to name it 'utils.ts', just for the sake of making the file trees shorter and easier to navigate

sg-gs (approved):

PR #135, opened 2022-09-26
[_]:(chore) Move login and register methods to use SDK instead of raw fetch

PixoDev:
Some methods in the app such login and register was using a raw fetch() to the drive server, this PR refactors those methods to use the internxt-sdk ones instead.


src/store/slices/auth/index.ts
```diff
@@ -58,12 +60,22 @@ export const initializeThunk = createAsyncThunk<void, void, { state: RootState }
   },
 );
 
+/**
+ * TO DO review the auth flow since this event driven stuff + Redux
```
sg-gs, line 64: Be more professional about comments/TODOs left on the code @PixoDev

sg-gs (approved):

PR #136, opened 2022-09-27
[_]:(feature) New complete survey referral step

PixoDev:
New referral step, complete survey.

Now the app refresh the storage and referral links on app focus when it comes back from background to foreground


src/services/NetworkService/realtimeUpdates.ts
```diff
@@ -1,33 +1,59 @@
...
-    const URL = `${constants.DRIVE_API_URL}`;
-    this.socket = io(URL, {
+  constructor(private sdk: SdkManager) {}
+
+  init() {
+    this.socket = io(constants.NOTIFICATIONS_URL, {
+      auth: {
+        token: this.sdk.getApiSecurity().photosToken,
+      },
       transports: ['websocket'],
```
sg-gs, line 15: remove this, as the we use the default transport

src/services/NetworkService/realtimeUpdates.ts
```diff
@@ -1,33 +1,59 @@
...
 
-  constructor() {
-    const URL = `${constants.DRIVE_API_URL}`;
-    this.socket = io(URL, {
+  constructor(private sdk: SdkManager) {}
+
+  init() {
+    this.socket = io(constants.NOTIFICATIONS_URL, {
+      auth: {
+        token: this.sdk.getApiSecurity().photosToken,
```
sg-gs, line 13: the naming is a bit confusing, is not a `photosToken`, is the new auth token with the proper format.

sg-gs (changes requested):

src/services/NetworkService/realtimeUpdates.ts
PixoDev: I know, but the name was already this when I started on mobile, is on the list of pending refactors, is maintained for compatibility and consistency, but will be definitely refactorized in the near future

src/services/NetworkService/realtimeUpdates.ts
PixoDev: doesnt work well on mobile, seems like it defaults to xhr polling, and it throws connection errors

sg-gs (approved):

PR #139, opened 2022-09-30
[_]:(hotfix) Fix incorrect sorting for drive items

PixoDev:
Some items were being sorted incorrectly in Drive.

What changed:

- The size being received was an string, so sorting by size was incorrect
- The name of the files has been converted to hex, so we can sort them in any language

### DIFF
--- a/src/screens/DriveScreen/index.tsx
+++ b/src/screens/DriveScreen/index.tsx
@@ -52,11 +52,8 @@ function DriveScreen({ navigation }: TabExplorerScreenProps<'Drive'>): JSX.Eleme
   const driveSortedItems = useMemo(
     () => [
       ...driveUploadingItems,
-      ...driveItems.sort(drive.file.getSortFunction(sortMode)).sort((a, b) => {
-        const aValue = a.data.fileId ? 1 : 0;
-        const bValue = b.data.fileId ? 1 : 0;
-        return aValue - bValue;
-      }),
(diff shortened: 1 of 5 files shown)


PixoDev:
https://github.com/internxt/drive-server/blob/master/src/app/models/file.ts#L41

As @JoanVicens  pointed the output from this can be an string, I verified that in mobile an string is received

src/screens/DriveScreen/index.tsx
```diff
@@ -52,11 +52,8 @@ function DriveScreen({ navigation }: TabExplorerScreenProps<'Drive'>): JSX.Eleme
   const driveSortedItems = useMemo(
     () => [
       ...driveUploadingItems,
-      ...driveItems.sort(drive.file.getSortFunction(sortMode)).sort((a, b) => {
-        const aValue = a.data.fileId ? 1 : 0;
-        const bValue = b.data.fileId ? 1 : 0;
-        return aValue - bValue;
-      }),
+      ...driveItems.filter((item) => !item.data.fileId).sort(drive.file.getSortFunction(sortMode)),
```
sg-gs, line 55: use `concat`

sg-gs, line 55: use concat

sg-gs (changes requested):

src/screens/DriveScreen/index.tsx
PixoDev: Solved

PR #144, opened 2022-10-18
[WT-145] feature/Upload file plain_name and use new types of files and folders

CandelR:


src/types/drive.ts
```diff
@@ -131,13 +131,16 @@ export interface SqliteDriveItemRow {
   icon_id: number | null;
   is_folder: boolean;
   name: string;
+  plain_name: string | null;
```
PixoDev, line 134: Why this can be null? Not all files contains plain_name? Or is because current SQlite stored files doesn't have one?

CandelR: Correct, I have left the possibility of null for files that are already stored and have no plain_name to avoid unexpected crashes.

PixoDev (approved):

src/services/drive/database/driveLocalDB.ts
```diff
@@ -137,9 +138,12 @@ class DriveLocalDB {
         folderId: row.parent_id as number,
         folder_id: row.parent_id as number,
         name: row.name,
+        plain_name: row.plain_name,
```
sg-gs, line 141: please use camelCase always if is not a limitation (SQLite etc)

sg-gs (changes requested):

src/services/drive/database/driveLocalDB.ts
CandelR: Okay! 
Here I have used it this way because it is the SDK typing, because from what I saw when getting the columns from the Postgres tables you get the field with the name as it is in the column and there is nothing to do a parsing, that's why they are in snake case. 
If I'm wrong about this I can check it and modify it!

PR #161, opened 2022-11-29
[MT-80]:(feature) Refresh tokens system

PixoDev:
This PR implements a system that refresh the auth tokens on app startup

1. Check if we have a token locally, if we have a token locally, decode it, get the expiration, if the expiration is less than Date.now() -> logout, else let the user access the app content
2. Refresh the auth tokens agains drive-server-wip

If step 1 or 2 throws an error, the user is logged out


src/services/AuthService.ts
```diff
@@ -29,6 +29,12 @@ export interface AuthCredentials {
   user: UserSettings;
 }
 
+export type JWTPayload = {
```
JoanVicens, line 32: jwt-decode also exports an interface called JwtPayload with some common jwt claims
```ts
export interface JwtPayload {
  iss?: string;
  sub?: string;
  aud?: string[] | string;
  exp?: number;
  nbf?: number;
  iat?: number;
  jti?: string;
}
```

sg-gs (approved):

PR #163, opened 2022-12-02
[_]: Added compatibility for Android 31

PixoDev:


src/services/drive/database/driveLocalDB.ts
```diff
@@ -134,7 +133,7 @@ class DriveLocalDB {
     return sqliteService.executeSql(DRIVE_DB_NAME, folderRecordTable.statements.deleteById, [folderId]);
   }
 
-  public async deleteItem({ id, isFolder }: { id: number; isFolder: boolean }) {
+  public async deleteItem({ id }: { id: number; isFolder: boolean }) {
```
fabioespinosa, line 136: Maybe remove isFolder from the type as well 🤔

fabioespinosa (changes requested):

src/services/drive/database/driveLocalDB.ts
PixoDev: Done

fabioespinosa (approved):

PR #164, opened 2022-12-02
[MT-78]: Expose .npmrc and use CI NPM_TOKEN env var

PixoDev:
Allow to run unit tests from Github Actions

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@ styles/tailwind.json
 .env.test.json
 .env.production.json
 .env.staging.json
-.npmrc
 google-services.json
 GoogleService-Info.plist
 
--- a/.github/workflows/lint-and-unit-tests.yml
+++ b/.github/workflows/lint-and-unit-tests.yml
@@ -1,8 +1,7 @@
-name: Staging CI
+name: Mobile - Lint and Unit Tests
 on:
   push:
-    # TMP for debugging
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -18,7 +18,6 @@ styles/tailwind.json
 .env.test.json
 .env.production.json
 .env.staging.json
-.npmrc
```
fabioespinosa, line 21: why are we removing to ignore .npmrc ?

fabioespinosa (changes requested):
I think its a better idea to use this approach:

https://github.com/internxt/drive-web/blob/master/.github/workflows/cd-to-staging.yml#L28

Inserting all the .npmrc file line by line in github actions than to removing the ignore from the .gitignore

This way we are always sure that .npmrc will never be uploaded

.npmrc.template
```diff
@@ -1,5 +0,0 @@
-registry=https://registry.yarnpkg.com/
-  
-@internxt:registry=https://npm.pkg.github.com
```
fabioespinosa, line 3: I don’t think it’s necessary to remove the .npmrc.template, could be useful for new devs 🤔

fabioespinosa (changes requested):

fabioespinosa (approved):

PR #166, opened 2022-12-12
[MT-51]:(chore) Upgrade RN dependencies for iOS

PixoDev:
Insights from this, worth a read:

- react-native-crypto should die at some point, is giving us a lot of dependency problems (mostly because is dead since 3 years ago)
- Expo upgrade solved a lot of problems, but still need to check the new publish system which is eas update https://docs.expo.dev/eas-update/introduction/

Some tweaks added:

- rn-nodeify is replaced by node-libs-react-native which is a better approach to mock NodeJS modules (we still should not use NodeJS modules) https://www.npmjs.com/package/node-libs-react-native

### DIFF
--- a/ios/.xcode.env.local
+++ b/ios/.xcode.env.local
@@ -1,1 +1,0 @@
-export NODE_BINARY="/Users/robert/.nvm/versions/node/v16.16.0/bin/node"
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ DerivedData
 *.ipa
 *.xcuserstate
 project.xcworkspace
+ios/.xcode.env.local
 
 # Android/IntelliJ
 android/app/release


ios/.xcode.env.local
```diff
@@ -0,0 +1 @@
+export NODE_BINARY="/Users/robert/.nvm/versions/node/v16.16.0/bin/node"
```
sg-gs, line 1: this is intentional?

src/network/crypto.ts
```diff
@@ -25,7 +25,7 @@ export async function decryptFileFromFs(
   },
 ): Promise<void> {
   const decipher = getAes256CtrDecipher(fileDecryptionKey.slice(0, 32), iv);
-  const fileSize = parseInt((await stat(fromPath)).size);
+  const fileSize = (await stat(fromPath)).size;
```
sg-gs, line 28: last time I was working with this codebase, the size field was a number stringified, is not the case anymore?

sg-gs (approved):

src/network/crypto.ts
PixoDev: Not anymore since react-native-fs is updated too

https://github.com/itinance/react-native-fs/blob/master/index.d.ts#L23

ios/.xcode.env.local
PixoDev: nope, excluding it

PR #168, opened 2022-12-15
[MT-86]:(feature) Drive file cache manager

PixoDev:
Implemented File cache manager, a system that given a directory maintains only the most recently used files that doesn't exceeds the given configurated limits

### DIFF
--- a/src/services/common/filesystem/fileCacheManager.spec.ts
+++ b/src/services/common/filesystem/fileCacheManager.spec.ts
@@ -1,0 +1,195 @@
+import fileSystemService from '@internxt-mobile/services/FileSystemService';
+import { ReadDirItem } from 'react-native-fs';
+import { FileCacheManager, FileCacheManagerConfig } from './fileCacheManager';
+
+jest.mock('@internxt-mobile/services/FileSystemService', () => {
+  return {
+    exists: jest.fn(),
+    readDir: jest.fn(),
+    getDirSize: jest.fn(),
+    statRNFS: jest.fn(),
+    moveFile: jest.fn(),
+    touch: jest.fn(),
+    unlinkIfExists: jest.fn(),
+  };
+});
+
+const mockedFs = jest.mocked(fileSystemService);
+const fileCacheManagerFactory = (config?: Partial<FileCacheManagerConfig>) =>
+  new FileCacheManager({
+    directory: '/cache',
+    maxFileSizeToCacheInBytes: 1024 * 1024 * 50,
+    maxSpaceInBytes: 1024 * 1024 * 500,
+    ...config,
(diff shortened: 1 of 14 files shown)


src/services/common/filesystem/fileCacheManager.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+    it('Should fail if the cache manager is not initialized', () => {
+      sut.cacheFile('/non_cached', 'hi.png').catch((err) => {
+        expect(err).toBeTruthy();
+      });
+    });
+
+    it('Should fail if the directory does not exists', async () => {
+      mockedFs.exists.mockImplementationOnce(async () => false);
+      mockedFs.getDirSize.mockImplementationOnce(async () => 1);
+      sut.init().catch((err) => expect(err).toBeTruthy());
```
sg-gs, line 41: is better to match it against the expected error, not just assume that the error thrown is the desired error. 

Therefore, create a class for each custom error of the cache, then check if the error is an instance of the expected error.

src/services/common/filesystem/fileCacheManager.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+describe('File Cache Manager', () => {
+  describe('When initializing the file cache manager', () => {
+    let sut: FileCacheManager = fileCacheManagerFactory();
+    beforeEach(() => {
+      sut = fileCacheManagerFactory();
+      jest.restoreAllMocks();
+    });
+    it('Should fail if the cache manager is not initialized', () => {
+      sut.cacheFile('/non_cached', 'hi.png').catch((err) => {
+        expect(err).toBeTruthy();
```
sg-gs, line 34: same here

src/services/common/filesystem/fileCacheManager.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+        mtime: Date.now(),
+        mode: 0,
+        path: '/non_cached/hi.png',
+        originalFilepath: '/non_cached/hi.png',
+        isFile: () => true,
+        isDirectory: () => false,
+      }));
+
+      const { cached } = await sut.cacheFile('/non_cached', 'hi.png');
+      expect(mockedFs.unlinkIfExists).toBeCalledTimes(2);
```
sg-gs, line 192: Could be improved a bit if the params for each call to `unlinkIfExists` is matched against the expected params. 

That match will virtually discard all side-effects that this test could face - like unlinking not the oldest files always, which will be a false positive case -.

sg-gs (changes requested):

sg-gs (approved):

PR #169, opened 2022-12-23
[_]:(fix) Identify files without extensions correctly

PixoDev:
Some files such .DS_Store or .env was being identified incorrectly as folders, the code flow should be improved for Drive in mobile since some interface mappings are not straightforward, but at least with this we can identify consistently a file and a folder

### DIFF
--- a/src/components/modals/SharedLinkSettingsModal/SharedLinkSettingsModal.tsx
+++ b/src/components/modals/SharedLinkSettingsModal/SharedLinkSettingsModal.tsx
@@ -138,12 +138,12 @@ export const SharedLinkSettingsModal: React.FC<SharedLinkSettingsModalProps> = (
 
     // If we already have a generated share link, copy it
     if (generatedShareLink) {
-      Clipboard.setString(generatedShareLink);
+      await Clipboard.setStringAsync(generatedShareLink);
 
       return;
     }
 
-    const isFolder = item?.type ? false : true;
+    const isFolder = item?.fileId ? false : true;
 
     // A share link already exists, obtain it
     if (item?.token && item?.code) {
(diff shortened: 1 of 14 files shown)


src/components/modals/SharedLinkSettingsModal/SharedLinkSettingsModal.tsx
```diff
@@ -138,12 +138,12 @@ export const SharedLinkSettingsModal: React.FC<SharedLinkSettingsModalProps> = (
...
     // If we already have a generated share link, copy it
     if (generatedShareLink) {
-      Clipboard.setString(generatedShareLink);
+      await Clipboard.setStringAsync(generatedShareLink);
 
       return;
     }
 
-    const isFolder = item?.type ? false : true;
+    const isFolder = item?.fileId ? false : true;
```
PixoDev, line 146: Should use isFolder property here, but is not common to all the interfaces we have. We have DriveItem, DriveListItem, DriveItemProps... is a bit messy and not all of them extends from the DriveItem base interface which makes this a nice refactor candidate

src/components/modals/SharedLinkSettingsModal/SharedLinkSettingsModal.tsx
```diff
@@ -138,12 +138,12 @@ export const SharedLinkSettingsModal: React.FC<SharedLinkSettingsModalProps> = (
 
     // If we already have a generated share link, copy it
     if (generatedShareLink) {
-      Clipboard.setString(generatedShareLink);
+      await Clipboard.setStringAsync(generatedShareLink);
```
PixoDev, line 141: Fixes some deprecation warnings since the other function blocks the thread for a moment since is synchronous and native (very bad idea)

sg-gs: Nice

sg-gs (approved):

PR #170, opened 2022-12-23
[_]:(fix) Avoid photo name collision in the local db

PixoDev:
Fixes a corner case where the same Photo can be inserted in the db twice, happens when:

1. We begin the sync process
2. We pull the photo from the server, but the photo in the device is already in the sync queue
3. The photo is already in the local db, but we check it agains the server since when we checked before against the local db, the Photo was not yet in the local DB
4. Photos server says the photo exists and gives us the Photo
5. We try to insert the Photo, but is alreay in the localDB causing a constraint violation in SQlite (not anymore with this PR)

### DIFF
--- a/src/services/photos/database/photosLocalDB.ts
+++ b/src/services/photos/database/photosLocalDB.ts
@@ -119,26 +119,14 @@ export class PhotosLocalDB {
 
   public async savePhotosItem(photo: Photo) {
     this.log('Saving photos item');
-    const photoExists = await this.getSyncedPhotoByPhotoId(photo.id);
-    if (photoExists) {
-      this.log('Photos item already exists, skipping insert');
-      await sqliteService.executeSql(PHOTOS_DB_NAME, deviceSyncTable.statements.updatePhotoById, [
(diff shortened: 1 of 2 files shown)


sg-gs:
Does this replacement affect the metadata in any use case where the photo is somewhat updated but the old version of the photo with old metadata persists in the local database?

PixoDev:
> Does this replacement affect the metadata in any use case where the photo is somewhat updated but the old version of the photo with old metadata persists in the local database?

We identify a Photo as unique in the backend using the PhotoID, but in mobile we don't have a PhotoID to match the device photos, so we use the name (thats why is unique in the table). As far as the PhotoID doesn't change, any update to a row with metadata changes (like the name, or the modifiedAt field) will update the row. This PR aims to fix the cases where there are a constraint conflict, but doesn't prevent us from updating the row.

Anyways right now, you can't update a Photo metadata, we would need a system to pull the "last updated photos" from the server, Mega uses an interesting system called sync heartbeat that ping the server and asks for updates if there's one

PR #171, opened 2022-12-27
[MT-77]:(feature) Enable/disable Photos sync from settings

PixoDev:
Enable or disable the Photos sync

### DIFF
--- a/src/contexts/Photos/Photos.context.tsx
+++ b/src/contexts/Photos/Photos.context.tsx
@@ -1,11 +1,14 @@
+import asyncStorageService from '@internxt-mobile/services/AsyncStorageService';
 import errorService from '@internxt-mobile/services/ErrorService';
 import photos from '@internxt-mobile/services/photos';
 import { photosLogger } from '@internxt-mobile/services/photos/logger';
 import { photosUtils } from '@internxt-mobile/services/photos/utils';
+import { AsyncStorageKey } from '@internxt-mobile/types/index';
 import { PhotosItem, PhotosSyncStatus } from '@internxt-mobile/types/photos';
 import * as MediaLibrary from 'expo-media-library';
 import React, { useEffect, useRef, useState } from 'react';
 import { DataProvider } from 'recyclerlistview';
+
 import { startSync } from './sync';
 
 export interface PhotosContextType {
@@ -13,6 +16,10 @@ export interface PhotosContextType {
   uploadedPhotosItems: PhotosItem[];
   syncedPhotosItems: PhotosItem[];
   ready: boolean;
+  syncEnabled: boolean;
+  enableSync: (
+    enable: boolean,
+  ) => Promise<{ canEnable: boolean; enabled: boolean; permissionsStatus: MediaLibrary.PermissionStatus }>;
   start: () => Promise<void>;
   getPhotosItem: (name: string) => PhotosItem | undefined;
   selection: {
@@ -48,6 +55,14 @@ export const PhotosContext = React.createContext<PhotosContextType>({
     const r2Key = `${r2.name}-${r2.takenAt}`;
     return r1Key !== r2Key;
   }),
+  enableSync: async () => {
+    return {
+      canEnable: false,
+      enabled: false,
+      permissionsStatus: MediaLibrary.PermissionStatus.UNDETERMINED,
+    };
+  },
+  syncEnabled: true,
   uploadProgress: 0,
   refresh: () => Promise.reject('Photos context not ready'),
   removePhotosItems() {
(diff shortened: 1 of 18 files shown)


src/store/slices/auth/index.ts
```diff
@@ -168,11 +168,19 @@ export const refreshTokensThunk = createAsyncThunk<void, void, { state: RootStat
     try {
       const currentAuthToken = await asyncStorageService.getItem(AsyncStorageKey.PhotosToken);
       if (!currentAuthToken) throw new Error('Auth token not found');
-      const { newToken, token } = await authService.refreshAuthToken(currentAuthToken);
+      const refreshed = await authService.refreshAuthToken(currentAuthToken);
 
+      if (!refreshed) {
```
PixoDev, line 173: Not sure about this one, if we are unable to refresh the tokens (network request failed, 504 server error, or simply an error different than a 401) we don't logout the user since we assume the token can still be valid, otherwise this will logout us in cases such poor network conditions and being unable to communicate with the server. Anyways the old tokens are still "valid" based on their expirations.

So should we assume any API error should sign out the user or only the 401 ones?

src/contexts/Photos/Photos.context.tsx
```diff
@@ -131,12 +145,37 @@ export const PhotosContextProvider: React.FC = ({ children }) => {
...
 
+  async function checkShouldEnableSync() {
+    const permissionStatus = await permissions.getPermissionsStatus();
+
+    // No permissions, disable the sync
+    if (
+      permissionStatus === MediaLibrary.PermissionStatus.UNDETERMINED ||
+      permissionStatus === MediaLibrary.PermissionStatus.DENIED
+    ) {
+      setSyncEnabled(false);
```
sg-gs, line 168: I would set this flag on the thenable promise after calling `checkShouldEnableSync` as the method name does not suggest that the responsibility of the function is setting anything rather than just checking if the sync should be enabled or not. 

Therefore, you will be following SRP

src/contexts/Photos/Photos.context.tsx
```diff
@@ -131,12 +145,37 @@ export const PhotosContextProvider: React.FC = ({ children }) => {
...
+
+      return false;
+    }
+
+    const isEnabled = (await asyncStorageService.getItem(AsyncStorageKey.PhotosSyncEnabled)) === 'true';
```
sg-gs, line 173: Any change of the format of how the asyncStorageService stores if the photos sync is enabled or not would affect all the code that checks against a string. I suggest hiding this check behind a method from the asyncStorageService like:

`asyncStorageService.isPhotosSyncEnabled() -> boolean`

Thus you have more flexibility to change the format, the key name or anything else in the future.

src/contexts/Photos/Photos.context.tsx
```diff
@@ -188,11 +227,84 @@ export const PhotosContextProvider: React.FC = ({ children }) => {
...
+      setSyncEnabled(false);
+      // We only can save strings in the async storage
+      await asyncStorageService.saveItem(AsyncStorageKey.PhotosSyncEnabled, 'false');
+      return {
+        canEnable: false,
+        enabled: false,
+        permissionsStatus,
+      };
+    }
+    // We only can save strings in the async storage
```
sg-gs, line 247: Same here, hiding it behind the asyncStorageService could be a better idea

src/screens/PhotosPermissionsScreen/index.tsx
```diff
@@ -24,9 +45,14 @@ function PhotosPermissionsScreen({ navigation }: PhotosScreenProps<'PhotosPermis
   ];
 
   useEffect(() => {
-    photosCtx.permissions.getPermissionsStatus().then((permissions) => {
-      setPermissions(permissions);
-    });
+    photosCtx.permissions
+      .getPermissionsStatus()
+      .then((permissions) => {
+        setPermissions(permissions);
```
sg-gs, line 51: This is a good example of applying SRP, `getPermissions` gives the permissions and another function is responsible for handling that information in any use-case desired way.

src/store/slices/auth/index.ts
sg-gs: We should not, just 401.

sg-gs (changes requested):

src/contexts/Photos/Photos.context.tsx
```diff
@@ -131,12 +144,42 @@ export const PhotosContextProvider: React.FC = ({ children }) => {
     }),
   );
 
+  useEffect(() => {
+    checkShouldEnableSync()
```
sg-gs, line 148: For future references and in order to avoid code duplication, you can pipe promises to avoid callback hells:

`checkShouldEnableSync()
      .then((enabled) => {
        return asyncStorageService.savePhotosSyncIsEnabled(enabled)
      })
       .then(() => setSyncEnabled(enabled))
      .catch((error) => {
        reportError(error);
      });`

src/services/AsyncStorageService.ts
```diff
@@ -21,6 +21,14 @@ class AsyncStorageService {
     });
   }
 
+  async photosSyncIsEnabled() {
+    return (await this.getItem(AsyncStorageKey.PhotosSyncEnabled)) === 'true' ? true : false;
```
sg-gs, line 25: return the comparation directly: 
`return (await this.getItem(AsyncStorageKey.PhotosSyncEnabled)) === 'true'`

sg-gs (changes requested):

sg-gs (approved):

PR #173, opened 2023-01-11
[MT-99]:(feature) EAS update configuration

PixoDev:
Allow OTA updates using EAS services with expo updates

https://docs.expo.dev/eas-update/introduction/


ios/Internxt.xcodeproj/xcshareddata/xcschemes/Internxt.xcscheme
```diff
@@ -41,7 +41,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Debug"
+      buildConfiguration = "Release"
```
sg-gs, line 44: is this intentional?

ios/Internxt/AppDelegate.mm
```diff
@@ -0,0 +1,166 @@
+#import "AppDelegate.h"
```
sg-gs, line 1: Please add a link reference to this source code or documentation if its own-made if possible

sg-gs (approved):

ios/Internxt/AppDelegate.mm
PixoDev: this code is boilerplate only

ios/Internxt.xcodeproj/xcshareddata/xcschemes/Internxt.xcscheme
PixoDev: those changes are auto generated from `expo prebuild` command, is not intentional but is not something we can prevent

PR #174, opened 2023-01-17
[MT-90]:(feature) New Drive preview for images, videos and PDFs

PixoDev:
### **Description**

This PR contains updated UI and UX flow for Drive files downloading.

Related Figma Desing: https://www.figma.com/file/x7HdCqjvZGFiWZ9fKXb1vy/Drive-Mobile?node-id=5088%3A39988&t=nqx9L5qK8u3XlVH4-1

### Implementation

A new screen is pushed to the navigator once the file download starts, the new screen called DrivePreviewScreen shows:

- Basic information of the file (name, mtime...)
- Icon referencing the file type
- A progress bar with a progress message

Once the file is downloaded the UI can either:

- Show a "no preview associated" mode and suggest the user to open the file in another app
- Display the file preview. Allowed previews right now are Images (png, jpg, jpeg, heic), videos (mov, mp4, avi) and pdf files


src/hooks/useDriveItem.ts
```diff
@@ -65,6 +76,8 @@ const useDriveItem = (props: UseDriveItemProps) => {
     thunk.then(() => {
       setIsDisabled(false);
     });
+
+    navigation.navigate('DrivePreview');
```
sg-gs, line 80: Use enums for this instead of hardcoded strings. Therefore, in the future, if some alias, ID or something else changes, there is just one change required: on the enum

src/screens/drive/DriveFolderScreen/DriveFolderScreen.tsx
```diff
@@ -78,17 +78,31 @@ export function DriveFolderScreen({ navigation }: DriveScreenProps<'DriveFolder'
...
+        openFileViewer: false,
       }),
     );
     const downloadAbort = () => {
       thunk.abort();
     };
 
     drive.events.setDownloadAbort(downloadAbort);
+
+    navigation.navigate('DrivePreview');
```
sg-gs, line 90: same here

src/screens/drive/DrivePreviewScreen/DrivePreviewScreen.tsx
```diff
@@ -0,0 +1,248 @@
...
+  };
+
+  const handleOpenFileWith = async (filename: string, downloadedFilePath: string) => {
+    try {
+      await fs.shareFile({
+        title: filename,
+        fileUri: downloadedFilePath,
+      });
+    } catch (error) {
+      errorService.reportError(error);
```
sg-gs, line 115: This reports just to Sentry or is notified via UI too?

src/screens/drive/DrivePreviewScreen/DrivePreviewScreen.tsx
```diff
@@ -0,0 +1,248 @@
...
+        />
+      );
+    }
+
+    // Render a video preview
+    if (downloadingFile.downloadedFilePath && hasVideoPreview && generatedThumbnail) {
+      return <DriveVideoPreview thumbnail={generatedThumbnail?.path} source={downloadingFile.downloadedFilePath} />;
+    }
+
+    // Render a PDF preview
```
sg-gs, line 212: better use a boolean called: 
`const renderPdfPreview = downloadingFile.downloadedFilePath && hasPdfPreview`
in order to avoid comments as code should be as self-explanatory as possible

sg-gs (changes requested):

src/hooks/useDriveItem.ts
PixoDev: This is typed, is not an enum but the argument is restricted to keyof RootStackParamList, which means you can use another string but TS will complain, making an enum for this doesnt make sense for me since this is already typed and has restricted values

src/screens/drive/DrivePreviewScreen/DrivePreviewScreen.tsx
PixoDev: Just to Sentry, adding UI feedback

sg-gs (approved):

PR #175, opened 2023-01-18
[MT-94]:(feature) Photos remote sync manager implementation

PixoDev:
### Description

This PR adds a new system that takes care to sync the Photos we already have in the cloud with the user device. We do this in order to avoid asking the server if the photos are already synced.

More info can be found here: https://inxt.atlassian.net/browse/MT-94

### Implementation

- PhotosRemoteSyncManager
A system that takes care of pulling all the Photos. Each photo is first stored in the RealmDB, once is stored a preview check is made, if no local preview is found, a preview is downloaded.

Now before starting the localSync, the remoteSync needs to finish first, with this we achieve:

- No race conditions, since we know all the Photos that are already synced before starting the sync
- More feedback for the current sync stage


app.config.ts
```diff
@@ -34,14 +33,13 @@ const appConfig: ExpoConfig & { extra: AppEnv & { NODE_ENV: AppStage; RELEASE_ID
   assetBundlePatterns: ['**/*'],
   runtimeVersion: packageJson.version,
   ios: {
+    jsEngine: 'jsc',
```
sg-gs, line 36: we run on Android in Hermes and on IOS in JSC then, isn't it?

PixoDev: correct, at least for now, Hermes works very well with low end devices on Android

src/contexts/Photos/Photos.context.tsx
```diff
@@ -198,33 +207,26 @@ export const PhotosContextProvider: React.FC = ({ children }) => {
...
-    if (existsSynced) return;
-    syncedPhotosItems.current.push(photosItem);
 
+  async function onRemotePhotosSynced() {
+    const syncedPhotos = await photos.realm.getSyncedPhotos();
+    syncedPhotosItems.current = syncedPhotos.map((syncedPhoto) => photos.utils.getPhotosItem(syncedPhoto));
     setPhotosInLocalDB(syncedPhotosItems.current.length);
     const mergedPhotosItems = photosUtils.mergePhotosItems(devicePhotosItems.current.concat(syncedPhotosItems.current));
     setDataSource(
       new DataProvider(function (r1, r2) {
```
sg-gs, line 227: what is `r1`, what is `r2`?

src/contexts/Photos/Photos.context.tsx
```diff
@@ -319,34 +327,44 @@ export const PhotosContextProvider: React.FC = ({ children }) => {
...
+      setPhotosInRemote(totalPhotosInRemote);
       const mergedPhotosItems = photosUtils.mergePhotosItems(
         devicePhotosItems.current.concat(syncedPhotosItems.current),
       );
 
       setDataSource(
         new DataProvider(function (r1, r2) {
-          const r1Key = `${r1.name}-${r1.takenAt}`;
-          const r2Key = `${r2.name}-${r2.takenAt}`;
+          const r1Key = getListKey(r1);
```
sg-gs, line 359: same here

src/helpers/services.ts
```diff
@@ -17,3 +17,6 @@ export class NotInitializedServiceError extends Error {
     this.message = `${this.service} not initialized in ${where}, check the stacktrace for more info`;
   }
 }
+
+// Waits x ms before resolving, useful to avoid RN bridge message collapse
+export const sleep = async (sleepTimeInMs: number) => new Promise((resolve) => setTimeout(resolve, sleepTimeInMs));
```
sg-gs, line 22: `async` is not required here

src/services/photos/database/photosRealmDB.ts
```diff
@@ -0,0 +1,170 @@
...
+};
+export class PhotosRealmDB {
+  private _realm: Realm | null = null;
+  private logger: BaseLogger = new BaseLogger({
+    tag: 'PHOTOS_REALM_DB',
+    enabled: false,
+  });
+
+  async init() {
+    this._realm = await Realm.open({
```
sg-gs, line 54: private is enough to make it private, using the `_something` convention is just for pure-JS code which wants to make it private as until [ES2022](https://betterprogramming.pub/es2022-a-complete-guide-on-private-static-class-fields-methods-and-static-initialization-blocks-f12943a86b2d) you were not able to make something private in a class

src/services/photos/database/photosRealmDB.ts
```diff
@@ -0,0 +1,170 @@
...
+      deleteRealmIfMigrationNeeded: true,
+    });
+
+    this.logger.info('Photos Realm DB is ready');
+  }
+
+  clear() {
+    if (!this._realm) return;
+    this.realm.write(() => {
+      this.realm.deleteAll();
```
sg-gs, line 65: what if this fails?

src/services/photos/database/photosRealmDB.ts
```diff
@@ -0,0 +1,170 @@
...
+    });
+
+    fs.unlinkIfExists(this.realm.path);
+  }
+
+  private get realm() {
+    if (!this._realm) throw new Error('No connection open, call init() first');
```
sg-gs, line 72: If you are not doing something special before getting `this.realm` or this property is virtual, which is not, there is no need to make this private and provide a getter after. 

Also, if you have access to the underlying persistence technology, the coupling will appear for sure. I would suggest using the Repository pattern that we use on the server side which forces you to decouple the persistence layer (get, find, etc) from the concrete implementation (Realm, SQLite, etc)

src/services/photos/database/photosRealmDB.ts
```diff
@@ -0,0 +1,170 @@
...
+
+    return this._realm as Realm;
+  }
+
+  public async saveDevicePhotos(devicePhotos: Asset[]) {
```
sg-gs, line 77: `async` is not required here. Also, I suppose that you are not waiting for the write operation to finish here, intentionally.

src/services/photos/database/photosRealmDB.ts
```diff
@@ -0,0 +1,170 @@
...
+            devicePhotoJSON: JSON.stringify(devicePhoto),
+            takenAt: new Date(devicePhoto.creationTime),
+          },
+          Realm.UpdateMode.All,
+        );
+      });
+    });
+  }
+
+  public async savePhotosItem(photo: Photo, getQueryTime = false) {
```
sg-gs, line 93: `async` not required here too

src/services/photos/database/photosRealmDB.ts
```diff
@@ -0,0 +1,170 @@
...
+    return docs.map(this.parseObject);
+  }
+
+  private parseObject(object: Realm.Object) {
+    const serialized = object.toJSON() as PhotoDoc;
+    return JSON.parse(serialized.photoJSON) as Photo;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  private parseFirst(objects: Realm.Results<any>) {
```
sg-gs, line 164: using `unknown` will be more semantic and may allow you to remove the `eslint-disable` line

src/services/photos/network/PhotosNetworkManager.ts
```diff
@@ -48,28 +48,29 @@ export class PhotosNetworkManager implements RunnableService<PhotosNetworkManage
...
-            this.addOperation({
-              ...task,
-              retries: task.retries + 1,
-            });
-            next(err, null);
-          }
-        });
+      try {
+        const result = await this.processUploadOperation(task);
+        await sleep(100);
```
sg-gs, line 59: why?

src/services/photos/network/PhotosNetworkManager.ts
```diff
@@ -48,28 +48,29 @@ export class PhotosNetworkManager implements RunnableService<PhotosNetworkManage
   }
 
   private createQueue() {
-    const queue = async.queue<PhotosNetworkOperation, Photo | null, Error>((task, next) => {
+    const queue = async.queue<PhotosNetworkOperation, Photo | null, Error>(async (task, next) => {
```
sg-gs, line 51: Mixing callbacks and async is dangerous

src/services/photos/network/PhotosNetworkManager.ts
```diff
@@ -144,6 +147,11 @@ export class PhotosNetworkManager implements RunnableService<PhotosNetworkManage
    */
   public async processUploadOperation(operation: PhotosNetworkOperation): Promise<OperationResult | null> {
     this.onUploadStartCallback(operation.photosItem);
+    const stopIfAborted = () => {
+      if (this.isAborted) {
```
sg-gs, line 151: if is already aborted, why throw the abort error again? Is [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) available in RN? This pattern is browser-native and provides a signal handled internally to check if is already been aborted.

src/services/photos/sync/devicePhotosScanner/devicePhotosScanner.ts
```diff
@@ -73,13 +73,17 @@ export class DevicePhotosScannerService extends RunnableService<DevicePhotosScan
     this.updateStatus(DevicePhotosScannerStatus.PAUSED);
   }
 
-  public hasFinished() {
-    return this.status === DevicePhotosScannerStatus.COMPLETED;
+  public get hasFinished() {
```
sg-gs, line 76: nice getter usage

src/services/photos/sync/photosRemoteSyncManager/photosRemoteSyncManager.ts
```diff
@@ -0,0 +1,201 @@
...
+    } catch (err) {
+      this.logger.error(`Preview download failed ${(err as Error).message}`);
+
+      // Retry the preview download, or fail, we cannot wait for all the previews
+      // to be downloaded, since we don't save the failed ones, we just retry the task
+      // if no success, we will go to the next Photos page ignoring the failed ones
+      // Note that No preview is not the same as no photo in the view, the photo and the
+      // metadata will appear, but not the preview image. GalleryItem component has a mechanism
+      // to detect failing previews and retry the download
+      if (task.retries < MAX_PREVIEW_DOWNLOAD_RETRIES) {
```
sg-gs, line 95: Should we retry and stop the next previews to load or just re-queue this task again? A nice discussion can be made here. This failure can make the sync always slower if the same preview needs always all the retries.

src/services/photos/usage/photosUsage.service.ts
```diff
@@ -6,6 +6,11 @@ export class PhotosUsageService {
     const { usage } = await this.sdk.photos.photos.getUsage();
     return usage;
   }
+
+  // Get total count of photos
+  public async getCount() {
```
sg-gs, line 11: `async` not required here

sg-gs (changes requested):

sg-gs:
As we are changing the persistence layer of the sync mechanism, I am going to be more precise with the changes and how they are added to the codebase @PixoDev

src/services/photos/database/photosRealmDB.ts
PixoDev: null is returned

src/services/photos/sync/photosRemoteSyncManager/photosRemoteSyncManager.ts
PixoDev: retrys always go to the end of the queue, so they will be retried once we finish with the other ones

src/services/photos/database/photosRealmDB.ts
PixoDev: actually RealmDB is sync, so the async should be removed

src/services/photos/network/PhotosNetworkManager.ts
PixoDev: yes is available, should move this to `abortController.signal.aborted`

Could happen that when we abort the system, and empty the queue, a task will be still being processed, for that case this is used, since not all the operations support aborts. For example the generate preview operation cannot be aborted, but we can check if the system is aborted before commiting

src/contexts/Photos/Photos.context.tsx
PixoDev: row1, row2, bad naming I aggree.

This function compares "what is different" to detect if we should render again or not

src/services/photos/database/photosRealmDB.ts
sg-gs: Go then

src/services/photos/database/photosRealmDB.ts
sg-gs: Here too

src/services/photos/usage/photosUsage.service.ts
```diff
@@ -6,6 +6,11 @@ export class PhotosUsageService {
     const { usage } = await this.sdk.photos.photos.getUsage();
     return usage;
   }
+
+  // Get total count of photos
```
sg-gs, line 10: Remove this comment and change the naming if is not clear, otherwise use JSDoc syntax

sg-gs (changes requested):

sg-gs (approved):

PR #178, opened 2023-01-30
Fix/profile avatar caching

PixoDev:
Cache avatar image correctly so we don't need to download it every time


sg-gs (approved):
Are we refreshing the avatar just when the avatar UUID key changes? That will be the optimum strategy to do that as we are doing on the web app.

PR #181, opened 2023-03-01
[MT-111]:(fix) Fix not playable video previews

PixoDev:
Fix not playable videos for Drive previews.

In some cases the native video player is not able to reproduce the video, usually due to codec incompatibility. on iOS we use AVFoundation, on Android ExoPlayer.

**Solution**

We detect when we cannot load the video in the in app player, when that happens, and the user taps on the play button, we prompt the user the native OS video player

### DIFF
--- a/src/network/NetworkFacade.ts
+++ b/src/network/NetworkFacade.ts
@@ -84,7 +84,10 @@ export class NetworkFacade {
     const plainFilePath = filePath;
 
     const encryptedFilePath = fileSystemService.tmpFilePath(`${uuid.v4()}.enc`);
-
+    // Some Android devices doesn't create the file when encrypting, they wait for
+    // the file to exist, since we remove the file and the end anyways
+    // we create an empty one here
+    await fileSystemService.createEmptyFile(encryptedFilePath);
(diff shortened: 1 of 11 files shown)


src/network/NetworkFacade.ts
```diff
@@ -84,7 +84,10 @@ export class NetworkFacade {
     const plainFilePath = filePath;
 
     const encryptedFilePath = fileSystemService.tmpFilePath(`${uuid.v4()}.enc`);
-
+    // Some Android devices doesn't create the file when encrypting, they wait for
+    // the file to exist, since we remove the file and the end anyways
+    // we create an empty one here
+    await fileSystemService.createEmptyFile(encryptedFilePath);
```
sg-gs, line 90: If we have always the file before encrypting, why is this required?

sg-gs: I am not sure if I am understanding this comment properly

PixoDev: My bad, this is not needed indeed, we should check if the file exists before the decrypt process starts, fixing

sg-gs (approved):

PR #188, opened 2023-03-11
[MT-105]:(fix) Fix photos refresh while sync is running

PixoDev:
- Fix photos refresh attempting to restart the sync process

### DIFF
--- a/src/services/photos/sync/devicePhotosScanner/devicePhotosScanner.ts
+++ b/src/services/photos/sync/devicePhotosScanner/devicePhotosScanner.ts
@@ -118,7 +118,10 @@ export class DevicePhotosScannerService extends RunnableService<DevicePhotosScan
 
   private saveCachedDevicePhotos(assets: MediaLibrary.Asset[]) {
     assets.forEach((asset) => {
-      const name = asset.filename.split('.')[0];
+      let name = asset.filename;
+      if (asset.filename.includes('.')) {
+        name = asset.filename.split('.')[0];
+      }
 
       const assetKey = `${name}-${asset.creationTime.toString()}`;
 
--- a/src/services/photos/utils/photosUtils.ts
+++ b/src/services/photos/utils/photosUtils.ts
@@ -23,11 +23,15 @@ export class PhotosUtils {
    * @param filename Filename usually provided by expo-media-library
    * @returns The photo type, like jpg, png, heic
    */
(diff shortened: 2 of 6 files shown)


src/services/photos/sync/devicePhotosScanner/devicePhotosScanner.ts
```diff
@@ -118,7 +118,10 @@ export class DevicePhotosScannerService extends RunnableService<DevicePhotosScan
 
   private saveCachedDevicePhotos(assets: MediaLibrary.Asset[]) {
     assets.forEach((asset) => {
-      const name = asset.filename.split('.')[0];
+      let name = asset.filename;
+      if (asset.filename.includes('.')) {
+        name = asset.filename.split('.')[0];
```
sg-gs, line 123: Be careful with photos whose name has been edited as the name should be from the beginning until before the last point that exists on the string

src/services/photos/utils/photosUtils.ts
```diff
@@ -23,11 +23,15 @@ export class PhotosUtils {
    * @param filename Filename usually provided by expo-media-library
    * @returns The photo type, like jpg, png, heic
    */
-  public getPhotoType(filename: string) {
-    const parts = filename.split('.');
-    const extension = parts[parts.length - 1];
+  public getPhotoType(filename: string): string | 'unknown' {
+    if (filename.includes('.')) {
+      const parts = filename.split('.');
+      const extension = parts[parts.length - 1];
```
sg-gs, line 29: Nice handled here

sg-gs (approved):

sg-gs:
Added a comment if that could be useful to take into consideration @PixoDev

src/services/photos/sync/devicePhotosScanner/devicePhotosScanner.ts
PixoDev: this filename is the one that the library provides us (expo-media-library). On iOS the filename cannot be edited since the images doesn't have filename at all, only an ID, (expo-media-library generates a consistent name from that).

On Android, since the filesystem is basically open, we only get the items that the library provides us based on our filters (we ask for photos and videos). I did a test and if you rename a file on Android from .jpg to .mp3 the item no longer is picked by expo-media-library so we are safe there.

PR #189, opened 2023-03-13
[PB-107]: Add file logger to app

PixoDev:
Logs are dumped into a file called internxt_runtime_logs.txt now in production, users will be able to share with us this file. This will give us better picture about what's happening on the user device without digging into Sentry


sg-gs (approved):

sg-gs:
As a suggestion: decoupling the logger is a nice step, use an interface to be able to switch or use multiple out channels, similar to a Unix Pipe. Now you are using it for Sentry, but other tools, APIs, or use cases could arise from logging all the actions that are going on in the app.

PixoDev:
> As a suggestion: decoupling the logger is a nice step, use an interface to be able to switch or use multiple out channels, similar to a Unix Pipe. Now you are using it for Sentry, but other tools, APIs, or use cases could arise from logging all the actions that are going on in the app.

I believe this is the `transport` here: https://github.com/internxt/drive-mobile/pull/189/files#diff-b6a37b0d76abeca0c20a27cde73a0fc9f97c737faefd775ce17fb8d56f59f972R9

You mean being able to do this for each logger? currently the `transport` option allows to do this, but for all the loggers

PR #191, opened 2023-03-15
[MT-106]:(fix) Improve gallery initial load times

PixoDev:
**Improved gallery initial load time**
The initial load time has been improved, now we first initialize the gallery with some photos (100) and the photos we have already synced in our local db, this way no matter how many photos the user has, we cut the initial load time for the 100 first photos only

**Previews improvements**
This PR fixes some issues on iOS when displaying items in the grid, note that the previews path has changed, but the Image on error callback will regenerate the preview in the correct path.


src/services/ErrorService.ts
```diff
@@ -65,6 +65,7 @@ class ErrorService {
       this.logger.warn(message);
     }
 
+    console.error(message);
```
sg-gs, line 68: remove this

src/services/photos/utils/photosUtils.ts
```diff
@@ -121,13 +121,23 @@ export class PhotosUtils {
...
+    type,
+  }: {
+    name: string;
+    takenAt: number;
+    size: PhotoSizeType;
+    type: string;
+  }) {
     if (size === PhotoSizeType.Full) {
-      return `${PHOTOS_FULL_SIZE_DIRECTORY}/${name}${type === 'unknown' || !type ? '' : '.' + type}`;
+      return `${PHOTOS_FULL_SIZE_DIRECTORY}/${name}_${takenAt}${type === 'unknown' || !type ? '' : '.' + type}`;
```
sg-gs, line 136: I suggesting using something that is not user-modifiable as a UUID mapped to the local database, just to ensure that nor the date neither name changing gives problems

sg-gs (changes requested):

src/services/photos/utils/photosUtils.ts
sg-gs: Just a suggestion (not a part of the required changes)

src/services/ErrorService.ts
PixoDev: This is intended, I'll add a comment explaining why in the code

src/services/photos/utils/photosUtils.ts
PixoDev: the problem with that is that you will need then to maintain the relationship between db and filesystem, some things that are hard to do here:

- Detect why and when a file was removed, then you will need to update the db

The paths are made this way so we can quickly found them in the filesystem, if we use the UUID, we will need first to look in the db, then in the filesystem, if the item is not there, update the db and update the filesystem, which are many bridge trips just for this.

According to how Photos works, if a user renames a photo, or changes the takenAt, we understand is a new photo, the hash that makes that photo unique will change too, as that won't match the backed up photo.

sg-gs (approved):

PR #209, opened 2023-10-05
[PB-570]: Fix/invalidate file cache correctly

masterprog-cmd:
**PROBLEM:**
When an item is modified on another platform (renamed or moved to the trash), the old item still appears (with the old name or in the folder from which the item was deleted).

**FIX:**
Now, we check if the item has been modified and is cached, and if so, we remove the item from the local DB to replace it with the new one.

### DIFF
--- a/src/contexts/Drive/Drive.context.tsx
+++ b/src/contexts/Drive/Drive.context.tsx
@@ -7,16 +7,18 @@ import {
   FetchFolderContentResponseWithThumbnails,
   FolderContent,
   FolderContentChild,
+  GetModifiedFiles,
 } from '@internxt-mobile/types/drive';
 import asyncStorageService from '@internxt-mobile/services/AsyncStorageService';
 import { AsyncStorageKey } from '@internxt-mobile/types/index';
 import drive from '@internxt-mobile/services/drive';
-import _ from 'lodash';
+import _, { last } from 'lodash';
 import errorService from '@internxt-mobile/services/ErrorService';
 import { driveLocalDB } from '@internxt-mobile/services/drive/database';
 import { BaseLogger } from '@internxt-mobile/services/common';
-import { getHeaders } from 'src/helpers/headers';
-import { constants } from '@internxt-mobile/services/AppService';
+import { driveFileService } from '@internxt-mobile/services/drive/file';
+import { driveFolderService } from '@internxt-mobile/services/drive/folder';
+import moment from 'moment';
 
 type DriveFoldersTree = {
   [folderId: number]:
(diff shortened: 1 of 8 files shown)


src/contexts/Drive/Drive.context.tsx
```diff
@@ -77,6 +82,92 @@ export const DriveContextProvider: React.FC<DriveContextProviderProps> = ({ chil
...
+  const checkIfItemShouldBeUpdatedOrDeleted = async (folderContentFromDB: FolderContent) => {
+    // 1. Get the modified items from the server
+    const [modifiedFiles, modifiedFolders] = await Promise.all([
+      getModifiedItems({
+        itemsType: 'files',
+        updatedAt: new Date().toISOString(),
+        status: 'ALL',
+      }),
+      getModifiedItems({
+        itemsType: 'files',
```
PixoDev, line 124: you are getting the `files` twice

src/contexts/Drive/Drive.context.tsx
```diff
@@ -77,6 +82,92 @@ export const DriveContextProvider: React.FC<DriveContextProviderProps> = ({ chil
     return driveUseCases.getFolderContent({ folderId });
   };
 
+  const getModifiedItems = async ({
```
PixoDev, line 83: move this to driveFile.service and driveFolder.service to get each array of updated items, don't place this logic here

https://github.com/internxt/drive-mobile/blob/master/src/services/drive/file/driveFile.service.ts
https://github.com/internxt/drive-mobile/blob/master/src/services/drive/folder/driveFolder.service.ts

src/contexts/Drive/Drive.context.tsx
```diff
@@ -77,6 +82,92 @@ export const DriveContextProvider: React.FC<DriveContextProviderProps> = ({ chil
...
+      !parsedModifiedFiles ||
+      !parsedModifiedFolders ||
+      folderContentFromDB.children.length === 0 ||
+      folderContentFromDB.files.length === 0
+    ) {
+      return;
+    }
+
+    folderContentFromDB.files.forEach((file: DriveFileData) => {
+      const fileModified = parsedModifiedFiles?.find((item: DriveFileData) => item.id === file.id);
```
PixoDev, line 143: Reduce the complexity of this operation, instead of traverse the current folder content db, traverse the parsedModifiedFiles and remove the stuff you want from the folderContentDb.

You are going to have always less modified files, than files in the folder, so is not the same to traverse 100 items than to traverse an array of 10 modified files.

Same for folders

src/contexts/Drive/Drive.context.tsx
```diff
@@ -77,6 +82,92 @@ export const DriveContextProvider: React.FC<DriveContextProviderProps> = ({ chil
...
+      if (fileModified) {
+        if (fileModified.status === 'TRASHED' || fileModified.status === 'REMOVED') {
```
PixoDev, line 145:
```
if(file.status === "EXISTS") {
// Item is not deleted
} else {
// Item is deleted
}
```

This is simpler in case you want to simplify this

PixoDev (changes requested):

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,6 +48,52 @@ interface DriveContextProviderProps {
 const logger = new BaseLogger({
   tag: 'DRIVE_CONTEXT',
 });
+
+export const removeOldItemsFromLocalDB = async () => {
```
PixoDev, line 52: maybe getModifiedDriveItems or something like that could explain better what this function does?

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,6 +48,52 @@ interface DriveContextProviderProps {
...
+  const lastUpdatedAt = (await asyncStorageService.getItem(AsyncStorageKey.LastUpdatedAt)) || new Date().toISOString();
+
+  const [modifiedFiles, modifiedFolders] = await Promise.all([
+    driveFileService.getModifiedFiles({
+      updatedAt: lastUpdatedAt,
```
PixoDev, line 57: You need to save filesLastUpdatedAt and foldersLastUpdatedAt separatedly, as the most recent folder could be newer than the most recent file, and you will miss items

EDIT: Never mind, I saw below than only the most recent will be picked, cool solution

src/contexts/Drive/Drive.context.tsx
masterprog-cmd: The thing is that the function also remove the old cached item🤔

masterprog-cmd: I'll see what I can do

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,6 +48,52 @@ interface DriveContextProviderProps {
...
+      status: 'ALL',
+    }),
+    driveFolderService.getModifiedFolders({
+      updatedAt: lastUpdatedAt,
+      status: 'ALL',
+    }),
+  ]);
+
+  if (!modifiedFiles || !modifiedFolders) return;
```
PixoDev, line 66:
```
// Check for the array length, and check both
if (!modifiedFiles.length && !modifiedFolders.length) return;
```

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,6 +48,52 @@ interface DriveContextProviderProps {
...
+  const modifiedFoldersIds = modifiedFolders.map((folder) => folder.id);
+
+  modifiedFilesIds.forEach((fileId) => {
+    driveLocalDB.deleteItem({ id: fileId }).catch((error) => {
+      errorService.reportError(error);
+    });
+  });
+
+  modifiedFoldersIds.forEach((folderId) => {
+    driveLocalDB.deleteFolderRecord(folderId).catch((error) => {
```
PixoDev, line 78: Don't do this, instead wait for the operation to complete blocking the code execution here's why:

```
// This timeout represents an async funcion, that is not being awaited
setTimeout(() => {
   deleteFolderRecord()
}, 500)

// If we read and the delete has not finished, we will still get the not yet deleted item
getDriveLocalDBItems()

File: src/services/drive/file/driveFile.service.ts [hunk 9]
```diff
@@ -233,6 +236,38 @@ class DriveFileService {
...
+      const modifiedItems = await fetch(`${constants.DRIVE_NEW_API_URL}/files?${query}`, {
+        method: 'GET',
+        headers,
+      });
+
+      const parsedModifiedFiles = await modifiedItems.json();
+
+      return parsedModifiedFiles;
+    } catch (error) {
+      errorService.reportError(error);
```
PixoDev, line 267: Propagate the error instead of catching it in the service, this way we can add context if we need to the reported error

PixoDev (changes requested):

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,6 +48,51 @@ interface DriveContextProviderProps {
...
+
+  modifiedFoldersIds.forEach(async (folderId) => {
+    await driveLocalDB.deleteFolderRecord(folderId);
+  });
+
+  // Get the last updatedAt date from the modified files and folders
+  let lastUpdatedAtFromModifiedFiles;
+  let lastUpdatedAtFromModifiedFolders;
+
+  if (modifiedFiles.length) lastUpdatedAtFromModifiedFiles = _.maxBy(modifiedFiles, 'updatedAt')?.updatedAt;
```
masterprog-cmd, line 86: Should I remove this conditional since we check if the arrays are not empty before? @PixoDev

src/contexts/Drive/Drive.context.tsx
PixoDev: `getModifiedDriveItemsAndUpdateLocalCache` could be a better name then, but up to you

src/contexts/Drive/Drive.context.tsx
PixoDev: nope, 

`if (!modifiedFiles.length && !modifiedFolders.length) return;`

Here you are checking that both, files and folders has no length, meaning that if one of them has length, this will not be true, so you'll need to check in case some of the arrays has no length

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,10 +50,75 @@ interface DriveContextProviderProps {
...
+    errorService.reportError(error);
+    return [[], []];
+  });
+
+  if (!modifiedFiles || !modifiedFolders || (!modifiedFiles.length && !modifiedFolders.length)) return;
+
+  const modifiedFilesIds = modifiedFiles.map((file) => file.id);
+  const modifiedFoldersIds = modifiedFolders.map((folder) => folder.id);
+
+  modifiedFilesIds.forEach(async (fileId) => {
```
PixoDev, line 76: This doesn't work as you think, use a for of loop
https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,10 +50,75 @@ interface DriveContextProviderProps {
...
+    await driveLocalDB.deleteItem({ id: fileId });
+  });
+
+  modifiedFoldersIds.forEach(async (folderId) => {
```
PixoDev, line 80: Same as the above comment

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,10 +50,75 @@ interface DriveContextProviderProps {
...
+};
+
 export const DriveContextProvider: React.FC<DriveContextProviderProps> = ({ children, rootFolderId }) => {
   const [viewMode, setViewMode] = useState(DriveListViewMode.List);
   const [driveFoldersTree, setDriveFoldersTree] = useState<DriveFoldersTree>({});
   const [currentFolder, setCurrentFolder] = useState<FetchFolderContentResponseWithThumbnails | null>(null);
+  const currentFolderId = useRef<number | null>(null);
+
+  const handleAppStateChange = async (state: AppStateStatus) => {
+    if (state === 'active') {
```
PixoDev, line 106: `if(state === 'active' && currentFolderId.current)`

This simplifies the code readability

src/contexts/Drive/Drive.context.tsx
```diff
@@ -46,10 +50,75 @@ interface DriveContextProviderProps {
...
+      if (currentFolderId.current)
+        loadFolderContent(currentFolderId.current, { pullFrom: ['network'] }).catch((error) => {
+          errorService.reportError(error);
+        });
+    }
+  };
+
+  useEffect(() => {
+    const listener = appService.onAppStateChange(handleAppStateChange);
```
PixoDev, line 115: move this to a ref, this variable can get lost without being removed cause it only lives in this scope, if moving to a ref, you can make sure you only have one always

src/contexts/Drive/Drive.context.tsx
```diff
@@ -88,14 +157,23 @@ export const DriveContextProvider: React.FC<DriveContextProviderProps> = ({ chil
    * 3. Network
    *
    */
+
   const loadFolderContent = async (folderId: number, options?: LoadFolderContentOptions) => {
     const shouldPullFromCache = options?.pullFrom ? options?.pullFrom.includes('cache') : true;
     const shouldPullFromNetwork = options?.pullFrom ? options?.pullFrom.includes('network') : true;
+
+    // Check if the items have been modified from another platform
```
PixoDev, line 165: Use comments only when they add something extra to the reader, your code should be self explanatory, `getModifiedDriveItems` is a name good enough to understand what happens, how it happens is not needed to be defined here, instead that's inside the function definition.

You can also use TSDoc in the function if you want to add more info

PixoDev (changes requested):

PixoDev:
Works great @masterprog-cmd, great job, merging this

PR #210, opened 2023-10-05
[PB-1047] feature/Shared links are not working on mobile

CandelR:
- Removed Shared Links view
- Add new call to create public share links
- Removed Shared Links options

### DIFF
--- a/src/screens/HomeScreen/index.tsx
+++ b/src/screens/HomeScreen/index.tsx
@@ -12,6 +12,7 @@ import * as useCases from '@internxt-mobile/useCases/drive';
 import { SearchInput } from 'src/components/SearchInput';
 import errorService from '../../services/ErrorService';
 import { SharedFiles, SharedFolders } from '@internxt/sdk/dist/drive/share/types';
+
 enum HomeTab {
   Recents = 'recents',
   Shared = 'shared',
@@ -23,7 +24,6 @@ const HomeScreen = (): JSX.Element => {
   const [sharedItemsPage, setSharedItemsPage] = useState(1);
   const [shouldGetMoreSharedFiles, setShouldGetMoreSharedFiles] = useState(true);
   const [shouldGetMoreSharedFolders, setShouldGetMoreSharedFolders] = useState(true);
-  const [gettingSharedItems, setGettingSharedItems] = useState(false);
   const [sharedItems, setSharedItems] = useState<(SharedFiles & SharedFolders)[]>([]);
 
   const {
@@ -53,39 +53,44 @@ const HomeScreen = (): JSX.Element => {
 
   useEffect(() => {
     handleSharedLinksRefresh();
-    const unsubscribe = useCases.onSharedLinksUpdated(getSharedItems);
+    const unsubscribe = useCases.onSharedLinksUpdated(async () => {
+      getSharedItems();
+    });
 
     return unsubscribe;
   }, []);
 
   const handleSharedLinksRefresh = async () => {
     try {
-      setGettingSharedItems(true);
+      setSharedItems([]);
       setShouldGetMoreSharedFiles(true);
       setShouldGetMoreSharedFolders(true);
+      setSharedItemsPage(1);
       const result = await getSharedItems({ page: 0, shouldGetFiles: true, shouldGetFolders: true });
-      if (!result?.data?.items.length) return;
 
-      if (!result.data.hasMoreFiles) {
+      if (!result?.data?.hasMoreFiles) {
         setShouldGetMoreSharedFiles(false);
       }
 
-      if (!result.data.hasMoreFolders) {
+      if (!result?.data?.hasMoreFolders) {
         setShouldGetMoreSharedFolders(false);
       }
+
+      if (!result?.data?.items.length) return;
+
       setSharedItems(result.data.items);
     } catch (error) {
       errorService.reportError(error);
-    } finally {
-      setGettingSharedItems(false);
     }
   };
 
   const handleNextSharedItemsPage = async () => {
     try {
-      setSharedItemsPage(sharedItemsPage + 1);
+      if (shouldGetMoreSharedFiles || shouldGetMoreSharedFolders) {
+        setSharedItemsPage(sharedItemsPage + 1);
+      }
       const result = await getSharedItems({
-        page: sharedItemsPage + 1,
+        page: sharedItemsPage,
         shouldGetFiles: shouldGetMoreSharedFiles,
         shouldGetFolders: shouldGetMoreSharedFolders,
       });
@@ -131,8 +136,8 @@ const HomeScreen = (): JSX.Element => {
       screen: (
         <SharedScreen
           searchText={searchText}
-          isLoading={gettingSharedItems}
-          sharedLinks={sharedItems as unknown as []}
+          isLoading={sharedLoading}
+          sharedLinks={sharedItems}
           refreshSharedLinks={handleSharedLinksRefresh}
           onEndOfListReached={handleNextSharedItemsPage}
         ></SharedScreen>
--- a/src/screens/HomeScreen/index.tsx
+++ b/src/screens/HomeScreen/index.tsx
@@ -12,6 +12,7 @@ import * as useCases from '@internxt-mobile/useCases/drive';
 import { SearchInput } from 'src/components/SearchInput';
 import errorService from '../../services/ErrorService';
 import { SharedFiles, SharedFolders } from '@internxt/sdk/dist/drive/share/types';
+import { sleep } from '../../helpers/services';
 
 enum HomeTab {
   Recents = 'recents',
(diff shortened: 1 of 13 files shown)


src/screens/HomeScreen/index.tsx
```diff
@@ -77,9 +131,10 @@ const HomeScreen = (): JSX.Element => {
       screen: (
         <SharedScreen
           searchText={searchText}
-          isLoading={sharedLoading}
-          sharedLinks={sharedLinks}
-          refreshSharedLinks={() => refreshSharedLinks()}
+          isLoading={gettingSharedItems}
+          sharedLinks={sharedItems as unknown as []}
```
PixoDev, line 135: I bet we can found a better way to do this

src/screens/drive/SharedScreen/SharedScreen.tsx
```diff
@@ -66,33 +68,37 @@ export const SharedScreen: React.FC<SharedScreenProps> = ({
     }
 
     if (sharedLinksToRender.length > 0) {
-      return sharedLinksToRender.map((sharedLink, i) => {
+      return sharedLinksToRender.map((sharedLink, i: React.Key) => {
+        logger.info({ sharedLinkData: sharedLink });
```
PixoDev, line 72: Add some context to the log, if not, this is just a wild object in the console/log file

src/services/drive/share/driveShare.service.ts
```diff
@@ -62,18 +43,46 @@ class DriveShareService {
     return this.sdk.share.deleteShareLink(shareId);
   }
 
+  stopSharingItem({ itemUUID, itemType }: { itemUUID: string; itemType: 'folder' | 'file' }) {
```
PixoDev, line 46: we'll need to prepare this service for mobile AFS, nothing to do here, but take this into account.

Anyways, why itemUUID instead of just uuid?

src/services/drive/share/driveShare.service.ts
```diff
@@ -62,18 +43,46 @@ class DriveShareService {
...
+  }
+
+  async getSharedFolders({
+    page,
+    perPage,
+    orderBy,
+  }: {
+    page: number;
+    perPage: number;
+    orderBy?: 'views:ASC' | 'views:DESC' | 'createdAt:ASC' | 'createdAt:DESC';
```
PixoDev, line 71: isn't this type in the SDK?

src/services/drive/share/driveShare.service.ts
```diff
@@ -62,18 +43,46 @@ class DriveShareService {
...
+  }
+
+  async getSharedFiles({
+    page,
+    perPage,
+    orderBy,
+  }: {
+    page: number;
+    perPage: number;
+    orderBy?: 'views:ASC' | 'views:DESC' | 'createdAt:ASC' | 'createdAt:DESC';
```
PixoDev, line 83: same, isn't this in the SDK?

src/types/drive.ts
```diff
@@ -204,14 +206,14 @@ export interface DriveCurrentFolderContent {
...
-export type FolderContentChild = Omit<FolderChild, 'uuid'>;
+export type FolderContentChild = FolderChild;
 
 export type FolderContent = Omit<FetchFolderContentResponse, 'children'> & {
   children: FolderContentChild[];
 };
 
 export interface FetchFolderContentResponseWithThumbnails extends FolderContent {
-  files: (DriveFileData & { thumbnail?: DownloadedThumbnail })[];
+  files: (DriveFileData & { thumbnail?: DownloadedThumbnail; uuid?: string })[];
```
PixoDev, line 216: betted to add uuid to DriveFileData, we will need this in more places

src/useCases/drive/deleteShareLink.ts
```diff
@@ -8,8 +8,7 @@ import { sharedLinksUpdated } from './getShareLink';
 
 export const deleteShareLink = async ({ shareId, type }: { shareId: string; type: 'file' | 'folder' }) => {
   try {
-    const result = await drive.share.deleteShareLink({ shareId });
-
+    await drive.share.stopSharingItem({ itemUUID: shareId, itemType: type });
```
PixoDev, line 11: so shareId is actually the item uuid? I'm having a hard time understanding variable names here

src/useCases/drive/deleteShareLink.ts
```diff
@@ -19,7 +18,7 @@ export const deleteShareLink = async ({ shareId, type }: { shareId: string; type
     analytics.track(DriveAnalyticsEvent.SharedLinkDeleted, {
       type,
     });
-    return result;
+    return true;
```
PixoDev, line 21: you sure about this? are we going to assume that whatever 200 the backend returns (or simply not an error) is a successful operation?

src/useCases/drive/loadSharedItems.ts
```diff
@@ -1,21 +1,85 @@
...
+  shouldGetFiles: boolean;
+}): Promise<
+  UseCaseResult<{ items: (SharedFiles & SharedFolders)[]; hasMoreFiles: boolean; hasMoreFolders: boolean }>
+> => {
   try {
-    const response = await drive.share.getShareLinks();
+    const [sharedFolders, sharedFiles] = await Promise.all([
+      shouldGetFolders
+        ? drive.share.getSharedFolders({ page, perPage: ITEMS_PER_PAGE })
+        : Promise.resolve<ListAllSharedFoldersResponse>({
```
PixoDev, line 46: what's this? if this is some kind of placeholder, is better to return null, otherwise we will need to maintain the placeholder and add all the data the placeholder needs in the future

src/useCases/drive/loadSharedItems.ts
```diff
@@ -1,21 +1,85 @@
...
+              networkPass: '',
+              networkUser: '',
+            },
+            files: [],
+            folders: [],
+            token: '',
+          }),
+      shouldGetFiles
+        ? drive.share.getSharedFiles({ page, perPage: ITEMS_PER_PAGE })
+        : Promise.resolve<ListAllSharedFoldersResponse>({
```
PixoDev, line 57: same with this placeholder, better to use a nully value

src/services/drive/share/driveShare.service.ts
CandelR: Nothing special, it is clearer to me because these objects have a wide range of fields. If you see it redundant because the name of the function already indicates that it is an _item_ we can leave it in **uuid** and **type**.

src/services/drive/share/driveShare.service.ts
CandelR: Nope 😞

src/services/drive/share/driveShare.service.ts
CandelR: Nope 😞

src/types/drive.ts
CandelR: `DriveFileData` is a type from sdk, we need to add it to the sdk and adapt all apps types.

src/useCases/drive/deleteShareLink.ts
CandelR: No, I just didn't finish checking the changes I made and it remained unchanged.
I change it.

src/useCases/drive/deleteShareLink.ts
CandelR: What other case apart from 200 are we going to want to handle here? If it is unsuccessful it will give some other HTTP error code and show the user that it has not been deleted. 
Is there any other success case we want to handle in the frontend?

src/services/drive/share/driveShare.service.ts
PixoDev: is more like to maintain a property - variable relationship, I prefer to have `stopSharingFile` and `stopSharingFolder` rather than this in any case cause is more explicit about what this is going to do.

Note that we usually don't use the `DriveItem` concept, instead we use `DriveFile` and `DriveFolder` concepts, but I agree that this is not a widely used pattern at Internxt, so this one is up to you, but whatever you decide, let's keep it consistent.

src/screens/HomeScreen/index.tsx
```diff
@@ -41,17 +45,72 @@ const HomeScreen = (): JSX.Element => {
...
+
+      if (!result?.data?.hasMoreFolders) {
+        setShouldGetMoreSharedFolders(false);
+      }
+
+      if (!result?.data?.items.length) return;
+
+      setSharedItems(sharedItems.concat(result.data.items));
+    } catch (error) {
+      errorService.reportError(error);
```
PixoDev, line 110: thanks for caring

src/types/drive.ts
PixoDev: use an alias then, let's try to decouple the Internxt SDK types and the Mobile types, as they don't need to match, we will refactor this on AFS mobile

src/useCases/drive/deleteShareLink.ts
PixoDev: a 400 status code is a good example of that, or maybe even a 204

PixoDev (approved):

PR #243, opened 2024-12-04
[PB-3286,PB-3278] bugfix/Error using "Take a photo and upload" and app crashes download big files

CandelR:
- Update project version to new release
- Added logs to record possible errors when upload images or download files


src/store/slices/drive/index.ts
```diff
@@ -242,6 +243,7 @@ const downloadFileThunk = createAsyncThunk<
 
       trackDownloadSuccess();
     } catch (err) {
+      logger.error('downloadFileThunk ', JSON.stringify(err));
```
xabg2, line 246: Maybe it would be better to add a clearer message like `Error downloading a file` or `Error in downloadFileThunk` or something like that.

xabg2 (approved):

rafijv (approved):

PR #246, opened 2024-12-19
[PB-3278] : feature/download in chunks

CandelR:
- Added range download and native function to join downloaded chunks


src/network/NetworkFacade.ts
```diff
@@ -203,49 +245,98 @@ export class NetworkFacade {
...
       },
       async (_, key, iv) => {
         if (!encryptedFileURI) throw new Error('No encrypted file URI found');
 
         // Maybe we should save the expected hash and compare even if the file is cached
-        if (!encryptedFileIsCached) {
-          await downloadJob.promise;
-          const sha256Hash = await RNFS.hash(encryptedFileURI, 'sha256');
-          const receivedFileHash = ripemd160(Buffer.from(sha256Hash, 'hex')).toString('hex');
+        // if (!encryptedFileIsCached) {
```
larryrider, line 331: this is commented on pourpose, right?

src/services/FileSystemService.ts
```diff
@@ -29,6 +29,22 @@ class FileSystemService {
...
     await this.prepareTmpDir();
   }
+  public async deleteFile(files: string[]): Promise<void> {
+    try {
+      await Promise.all(
+        files.map(async (file) => {
+          try {
+            await this.unlinkIfExists(file);
+          } catch (error) {
+            console.warn(`Error deleting file ${file}:`, error);
```
larryrider, line 39: console.warn here

src/network/NetworkFacade.ts
CandelR: Correct, I've added a comment

PR #248, opened 2024-12-30
[PB-3582] bugfix/Thumbnails not showing

CandelR:
- Added download function in one part again for thumbnails

### DIFF
--- a/src/network/NetworkFacade.ts
+++ b/src/network/NetworkFacade.ts
@@ -159,7 +159,148 @@ export class NetworkFacade {
     return [wrapUploadWithCleanup(), () => null];
   }
 
-  download(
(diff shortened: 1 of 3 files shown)


src/network/NetworkFacade.ts
```diff
@@ -159,7 +159,148 @@ export class NetworkFacade {
     return [wrapUploadWithCleanup(), () => null];
   }
 
-  download(
+  download(fileId: string, bucketId: string, mnemonic: string, params: DownloadFileParams): [Promise<void>, Abortable] {
+    if (!fileId) {
+      throw new Error('Download error code 1');
```
xabg2, line 164: Use better errors

xabg2 (changes requested):

PR #249, opened 2024-12-31
[PB-3542, PB-3489]: bugfix/Download issues

CandelR:
- Improved cancel download logic to avoid download same file multiple times in Preview screen


src/store/slices/drive/DownloadAnalytics.ts
xabg2, line 1: Do we still using analytics?

xabg2 (approved):

PR #252, opened 2025-01-10
[PB-3578] feature/Increase download limit to 10GB per file

CandelR:
- Increased download limit to 10GB
- Added animation to dots in preview view in order to give more visual feedback to users


src/network/NetworkFacade.ts
```diff
@@ -320,18 +322,25 @@ export class NetworkFacade {
...
 
     const decryptFileFromFs: DecryptFileFromFsFunction =
       Platform.OS === 'android' ? androidDecryptFileFromFs : iosDecryptFileFromFs;
 
     let encryptedFileURI: string | undefined;
-
     const encryptedFileName = `${fileId}.enc`;
     let encryptedFileIsCached = false;
-    let totalBytes = 0;
+    const totalBytes = 0;
```
xabg2, line 343: Seems this variable is not used anymore.

ios/Internxt/Info.plist
```diff
@@ -33,7 +33,7 @@
       </dict>
     </array>
     <key>CFBundleVersion</key>
-    <string>11</string>
+    <string>5</string>
```
larryrider, line 36: from 11 to 5 is right?

CandelR: Yes, no problem. This field is self-managed and modified when it is uploaded to the appstore. Apart from the app version, there is the bundle version, e.g. 1.7.0 (1), 1.7.0.(2), etc.
Anyway I will change it to 1

xabg2 (approved):

ios/Internxt/Info.plist
CandelR: I merge this and I will change it in the release PR, in order to have it updated.

PR #255, opened 2025-01-16
[PB-3578] bugfix/Cannot upload files bigger than 2GB on Android 

CandelR:
- Added multipart upload
- Add tests


src/screens/SecurityScreen/index.tsx
```diff
@@ -118,7 +118,8 @@ const SecurityScreen = ({ navigation }: SettingsScreenProps<'Security'>) => {
               />
             ) : null}
             {/* CHANGE PASSWORD */}
-            <SettingsGroup
+            {/* TEMPORARY DISABLED */}
+            {/* <SettingsGroup
```
larryrider, line 122: is this intended? if so, it looks good to me

larryrider (approved):

PR #266, opened 2025-05-16
[PB-4324] feature/Migrate user related calls

CandelR:
- Migrated user related enpdoints to point server-wip backend

Note:
    - Need https://github.com/internxt/drive-mobile/pull/265 updated


src/store/slices/auth/index.ts
```diff
@@ -105,11 +105,8 @@ export const signInThunk = createAsyncThunk<
     newToken: payload.newToken,
   });
   if (!payload.user.root_folder_id) {
-    const initializedUser = await UserService.initializeUser(payload.user.email, payload.user.mnemonic);
-
     userToSave = {
       ...userToSave,
```
larryrider, line 109: If the initializedUser logic is not longer needed, then this line makes no sense 🤔 
userToSave = {
      ...userToSave,
}

larryrider (approved):

PR #273, opened 2025-06-05
[PB-178] Feature/Fix status bar colors

CandelR:
- Changed status bar colors for better visibility in android with white status bar icons
- Fixed load invoices and load subscription call with correct params

### DIFF
--- a/src/components/modals/BottomModal/index.tsx
+++ b/src/components/modals/BottomModal/index.tsx
@@ -35,8 +35,6 @@ const BottomModal = (props: BottomModalProps): JSX.Element => {
   const safeAreaColor = props.safeAreaColor || getColor('bg-surface');
 
   const statusBarStyle = isDark ? 'light' : 'dark';
-  const statusBarBackgroundColor =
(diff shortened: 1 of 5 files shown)


src/components/modals/CenterModal/index.tsx
```diff
@@ -77,7 +90,7 @@ const CenterModal = ({
       easing={Easing.ease}
     >
       <Animated.View style={[tailwind('h-full'), animatedStyle]}>
-        <StatusBar translucent />
+        <StatusBar style={statusBarStyle} translucent={Platform.OS === 'android'} />
```
xabg2, line 93: Same here.

src/components/modals/BottomModal/index.tsx
```diff
@@ -47,7 +51,7 @@ const BottomModal = (props: BottomModalProps): JSX.Element => {
       easing={Easing.ease}
     >
       <View style={tailwind('h-full')}>
-        <StatusBar translucent />
+        <StatusBar style={statusBarStyle} translucent={Platform.OS === 'android'} />
```
xabg2, line 54: Perhaps you can extract Platform.OS === 'android' to a constant like const shouldBeTranslucent = Platform.OS === 'android'.

PR #276, opened 2025-06-09
[PB-4521] bugfix/White screen after screen lock

CandelR:
- Updated cleanup logic to reset screen lock states when not enabled
- Refresh token when app goes foreground only when expiration date is less than 3 days

### DIFF
--- a/src/services/AuthService.ts
+++ b/src/services/AuthService.ts
@@ -315,6 +315,27 @@ class AuthService {
   }
 
   /**
+   * Checks if token needs refresh (expires in less than 3 days)
+   * @param authToken The token to check
+   * @returns true if token expires in less than 7 days
+   */
+  public tokenNeedsRefresh(authToken: string): boolean {
+    try {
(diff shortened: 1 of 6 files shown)


src/services/AuthService.ts
```diff
@@ -314,6 +314,27 @@ class AuthService {
...
+    try {
+      const payload = jwtDecode<JWTPayload>(authToken);
+
+      if (!payload.exp) return true;
+
+      const expiration = payload.exp * 1000;
+      const THREE_DAYS_IN_MS = 3 * 24 * 60 * 60 * 1000;
+      const threeDaysFromNow = Date.now() + THREE_DAYS_IN_MS;
+
+      return expiration < threeDaysFromNow;
```
larryrider, line 332: A thing to take into account is that if a session has expired, you cannot refresh the token, a re-login is needed.

This makes sense as you should not be able to get a valid token using an invalid token.

PR #283, opened 2025-06-19
[PB-4530-4531] feature/Prevent screen captures and recording

CandelR:
- Added `react-native-capture-protection` to protect from screenshots, screen recording and view the app when minimized.

Note ⚠️ :
- I'll add an option in Settings to disable or enable these options, as it can hurt us when try to record the app


larryrider (approved):
Check SonarCloud issues

PR #295, opened 2025-07-14
[PB-4557] bugfix/Items list not refresh correctly

CandelR:
**Bug:**
- UI not reflecting moved items in drive list correctly after move it 

**Fix applied:**
- Optimistic updates: Item move to folder and update UI and local state instantly
- Auto rollback: Reverts all changes if move item call to API fails

### DIFF
--- a/src/store/slices/drive/index.ts
+++ b/src/store/slices/drive/index.ts
@@ -460,7 +460,6 @@ export const driveSlice = createSlice({
       state.folderContent = action.payload;
     },
     setFocusedItem(state, action: PayloadAction<DriveItemFocused | null>) {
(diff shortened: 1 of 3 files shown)


src/store/slices/drive/index.ts
```diff
@@ -450,6 +460,7 @@ export const driveSlice = createSlice({
       state.folderContent = action.payload;
     },
     setFocusedItem(state, action: PayloadAction<DriveItemFocused | null>) {
+      console.log('Setting focused item:', action.payload);
```
larryrider, line 463: is this console.log needed?

larryrider (approved):

PR #296, opened 2025-07-15
[PB-4530] featur/Screen recording protection

CandelR:
- Addes provider to detect correctly the screen protection status


src/screens/SettingsScreen/index.tsx
```diff
@@ -49,7 +49,7 @@ function SettingsScreen({ navigation }: SettingsScreenProps<'SettingsHome'>): JS
   const scrollViewRef = useRef<ScrollView | null>(null);
 
   const [isDarkMode, setIsDarkMode] = useState(false);
-  const [screenProtectionEnabled, setScreenProtectionEnabled] = useState(protectionStatus?.screenshot ? false : true);
+  const [screenProtectionEnabled, setScreenProtectionEnabled] = useState(protectionStatus?.screenshot);
```
xabg2, line 52: Maybe a fallback could be added here? `?? false` by default, for example.

PR #305, opened 2025-08-26
[PB-4845] feature/Migrate Internxt Drive mobile from api.internxt.com to gateway

CandelR:
- Bump version to 1.8.2 
- Update sdk for add cloudflare token to be able to use cloudflare endpoints

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "@internxt/lib": "^1.2.0",
     "@internxt/mobile-sdk": "^0.2.41",
     "@internxt/rn-crypto": "0.1.15",
-    "@internxt/sdk": "1.11.00",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -36,7 +36,7 @@
     "@internxt/lib": "^1.2.0",
     "@internxt/mobile-sdk": "^0.2.41",
     "@internxt/rn-crypto": "0.1.15",
-    "@internxt/sdk": "1.10.09",
+    "@internxt/sdk": "1.11.00",
```
larryrider, line 39: fix it to 1.11.0, it should work the same but..

larryrider (approved):

PR #306, opened 2025-08-27
[_] bugfix/Fix appscreen layout that makes content surpass screen bounds

CandelR:
- Fixed appscreen layout that makes content surpass screen bounds


xabg2 (approved):

larryrider:
check sonarcloud issues

PR #311, opened 2025-10-01
[PB-4965] bugfix/".folder" naming issue

CandelR:
- Implement `getDisplayName` function and update `DriveNavigableItem` to fix issue that displays folder names with `.folder` extension as the files in move items modal

### DIFF
--- a/src/helpers/itemNames.ts
+++ b/src/helpers/itemNames.ts
@@ -25,7 +25,7 @@ import { DriveItemDataProps } from '../types/drive';
  * getDisplayName({ name: 'unknown', isFolder: false }) // 'unknown'
  */
 export const getDisplayName = (item: DriveItemDataProps): string => {
-  if (item.isFolder || !item.type) return item.name;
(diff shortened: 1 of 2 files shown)


src/helpers/itemNames.ts
```diff
@@ -17,25 +17,15 @@ import { DriveItemDataProps } from '../types/drive';
...
-  }
-
-  const normalizedExtension = item.type.trim().toLowerCase();
-  const normalizedName = item.name.toLowerCase();
-
-  const expectedEnding = `.${normalizedExtension}`;
-  if (normalizedName.endsWith(expectedEnding)) {
-    return item.name;
-  }
+  if (item.isFolder || !item.type) return item.name;
```
larryrider, line 28: check also if type is present but its an empty string

larryrider (changes requested):

src/helpers/itemNames.ts
larryrider: sorry, I mean blank string (with spaces)

PR #316, opened 2025-10-06
[PB-4966] bugfix/Fix screen protection state

CandelR:
- Fix screen protection switch state to display correct one
- Moved logic to handle protection state to a hook
- Added tests


xabg2 (approved):
LGTM

PR #395, opened 2026-03-27
[PB-6187] feature/Move large files upload to main app

CandelR:
Main changes:
- **Native handoff for large files (>300 MB)**: When the iOS share extension detects files exceeding the 300 MB threshold, it now writes `pending_share_upload.json` to the iOS App Group and opens the main app via the `handle-large-share` deep link, bypassing the share extension JS entirely.
- **`LargeShareUploadScreen`**: New screen in the main app that reads the pending share payload from the App Group, reuses `DriveScreen` + `useShareUpload` for folder picking and upload, and skips the size check (`skipSizeCheck: true`) since the file already passed the Swift-layer validation.
- **`AppGroupPendingShareService`**: New native module bridge (`AppGroupPendingShareModule.swift`) to read/clear `pending_share_upload.json` from the shared App Group container.
- **`PHAssetExportModule`**: New native module to export Photos-library assets (`ph://` URIs) to a temp file path before upload (required for files picked from the Photos app).
- **`useDeepLinks`**: Extracted deep-link handling out of `RootNavigator` into a dedicated hook; extended to handle `handle-large-share` links and route to `LargeShareUploadScreen`.
- **`useIosPendingShareHandoff`**: New hook that checks on app foreground if a pending large share exists and navigates to `LargeShareUploadScreen`.


ios/Internxt.xcodeproj/xcshareddata/xcschemes/Internxt.xcscheme
```diff
@@ -41,7 +41,7 @@
       </Testables>
    </TestAction>
    <LaunchAction
-      buildConfiguration = "Release"
+      buildConfiguration = "Debug"
```
larryrider, line 44: is this intentional?

larryrider (approved):

ios/Internxt.xcodeproj/xcshareddata/xcschemes/Internxt.xcscheme
CandelR: Yes :)

PR #399, opened 2026-03-31
[PB-6231]  feature/Handle file name collisions before upload

CandelR:
Main changes: 
- Implemented name collision handling and modal for file uploads
- Apply URL decoding (decodeURIComponent) to shared file names to resolve percent-encoded characters (e.g. my%20file.pdf → my file.pdf) passed by the iOS system share sheet.

### DIFF
--- a/src/shareExtension/hooks/useShareExtension.ios.ts
+++ b/src/shareExtension/hooks/useShareExtension.ios.ts
@@ -37,19 +37,19 @@ export const useShareExtension = ({
       ...(files ?? []).map((uri) => ({
         uri,
         mimeType: getMimeTypeFromUri(uri),
-        fileName: uri.split('/').pop() ?? null,
+        fileName: decodeURIComponent(uri.split('/').pop() ?? '') || null,
         size: readSize(uri),
       })),
       ...(images ?? []).map((uri) => ({
         uri,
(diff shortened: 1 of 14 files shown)


src/shareExtension/hooks/useShareExtension.ios.ts
```diff
@@ -37,19 +37,19 @@ export const useShareExtension = ({
       ...(files ?? []).map((uri) => ({
         uri,
         mimeType: getMimeTypeFromUri(uri),
-        fileName: uri.split('/').pop() ?? null,
+        fileName: decodeURIComponent(uri.split('/').pop() ?? '') || null,
```
larryrider: preffer `??` over `||`

larryrider (approved):

src/shareExtension/hooks/useShareExtension.ios.ts
CandelR: Not here because `'' `is not nullish, which is needed for pass to the null value if we use ??

TamaraFinogina (approved):

src/shareExtension/hooks/useShareExtension.ios.ts
TamaraFinogina: btw, it repeats 3 times, maybe it can be a variable or computed before

PR #422, opened 2026-04-20
[PB-5925]: add native InternxtApiClient for DocumentsProvider

terrerox:
Adds a pure-Kotlin HTTP client that talks directly to the Drive and Bridge APIs so the SAF DocumentsProvider can fetch folder contents and download links without the React Native bridge. Covers list/create/rename/move/trash on Drive (Bearer auth) and getDownloadLinks on Bridge (Basic auth with sha256(userId).hex derivation). Typed errors for 401/404/other/network; MockWebServer tests for parsing, auth headers, gateway-required headers, and error mapping.

### DIFF
--- a/android/app/src/main/java/com/internxt/cloud/documents/api/AuthConfig.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/api/AuthConfig.kt
@@ -6,7 +6,7 @@ data class AuthConfig(
     val bearerToken: String,
     val bridgeUser: String,
     val userId: String,
-    val clientName: String = "drive-mobile",
-    val clientVersion: String = "v1.9.0",
+    val clientName: String,
+    val clientVersion: String,
     val desktopToken: String? = null
 )
--- a/android/app/src/main/java/com/internxt/cloud/documents/api/AuthConfig.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/api/AuthConfig.kt
@@ -1,0 +1,12 @@
+package com.internxt.cloud.documents.api
+
+data class AuthConfig(
+    val driveBaseUrl: String,
+    val bridgeBaseUrl: String,
+    val bearerToken: String,
+    val bridgeUser: String,
+    val userId: String,
+    val clientName: String = "drive-mobile",
+    val clientVersion: String = "v1.9.0",
+    val desktopToken: String? = null
+)
--- a/android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
@@ -192,8 +192,7 @@ class InternxtApiClient(
         private val JSON = "application/json; charset=utf-8".toMediaType()
 
         private fun defaultClient(): OkHttpClient = OkHttpClient.Builder()
-            .connectTimeout(30, TimeUnit.SECONDS)
-            .readTimeout(30, TimeUnit.SECONDS)
+            .callTimeout(30, TimeUnit.SECONDS)
             .build()
     }
 }
--- a/android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
@@ -44,7 +44,7 @@ class InternxtApiClient(
             .addQueryParameter("sort", "plainName")
             .addQueryParameter("order", "ASC")
             .build()
-        val body = execute(driveRequest(url).get().build())
+        val body = executeApiRequest(driveRequest(url).get().build())
         return body.optJSONArray(jsonKey).orEmpty().map(parse)
     }
 
@@ -55,7 +55,7 @@ class InternxtApiClient(
         val req = driveRequest(driveUrl("folders"))
             .post(payload.toString().toRequestBody(JSON))
             .build()
-        return parseFolder(execute(req))
+        return parseFolder(executeApiRequest(req))
     }
 
     fun renameFile(fileUuid: String, newName: String): DriveFile {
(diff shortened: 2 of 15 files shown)


android/app/src/main/java/com/internxt/cloud/documents/api/AuthConfig.kt
```diff
@@ -0,0 +1,12 @@
+package com.internxt.cloud.documents.api
+
+data class AuthConfig(
+    val driveBaseUrl: String,
+    val bridgeBaseUrl: String,
+    val bearerToken: String,
+    val bridgeUser: String,
+    val userId: String,
+    val clientName: String = "drive-mobile",
+    val clientVersion: String = "v1.9.0",
```
CandelR: Perhaps we should make this configurable so that there is a single source of truth. Otherwise, we might as well forget about changing it here

android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
```diff
@@ -0,0 +1,216 @@
...
+            .header("internxt-client", config.clientName)
+            .header("internxt-version", config.clientVersion)
+        config.desktopToken?.let { builder.header("x-internxt-desktop-header", it) }
+        return builder
+    }
+
+    private fun driveUrl(path: String) = "${config.driveBaseUrl.trimEnd('/')}/$path".toHttpUrl()
+    private fun bridgeUrl(path: String) = "${config.bridgeBaseUrl.trimEnd('/')}/$path".toHttpUrl()
+
+    private fun execute(request: Request): JSONObject {
```
CandelR: I think that this fun name could be a little more descriptive

android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
```diff
@@ -0,0 +1,216 @@
...
+}
+
+private fun JSONObject.optStringOrNull(key: String): String? =
+    if (isNull(key)) null else optString(key).takeIf { it.isNotEmpty() }
+
+private fun JSONObject.optLongFlexible(key: String): Long = when (val v = opt(key)) {
+    is Number -> v.toLong()
+    is String -> v.toLongOrNull() ?: 0L
+    else -> 0L
+}
```
CandelR: Perhaps I would extract these functions and add some tests

android/app/src/main/java/com/internxt/cloud/documents/crypto/HashUtil.kt
```diff
@@ -0,0 +1,21 @@
+package com.internxt.cloud.documents.crypto
+
+import java.security.MessageDigest
+
+object HashUtil {
```
CandelR, line 5: add @TamaraFinogina  as codeowner for crypto directory please :)

android/app/src/test/java/com/internxt/cloud/documents/api/InternxtApiClientTest.kt
```diff
@@ -0,0 +1,155 @@
...
+import okhttp3.mockwebserver.MockResponse
+import okhttp3.mockwebserver.MockWebServer
+import okhttp3.mockwebserver.SocketPolicy
+import org.junit.After
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertThrows
+import org.junit.Before
+import org.junit.Test
+
+class InternxtApiClientTest {
```
CandelR, line 14: I would add more tests, for example for listFolderFolders, createFolder, null optional fields in parseFile/parseFolder, size as a string value, and 5xx ApiError

android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
```diff
@@ -0,0 +1,199 @@
...
+            .put("parentFolderUuid", parentUuid)
+        val req = driveRequest(driveUrl("folders"))
+            .post(payload.toString().toRequestBody(JSON))
+            .build()
+        return parseFolder(executeApiRequest(req))
+    }
+
+    fun renameFile(fileUuid: String, newName: String): DriveFile {
+        val payload = JSONObject().put("name", newName)
+        val req = driveRequest(driveUrl("files/$fileUuid"))
```
CandelR, line 63: I think that the correct endpoint for rename is files/$fileUuid/meta and is a put, not a patch, unless there is a new one. Confirm it please

android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    companion object {
+        const val DEFAULT_PAGE_SIZE = 50
+
+        private val JSON = "application/json; charset=utf-8".toMediaType()
+
+        private fun defaultClient(): OkHttpClient = OkHttpClient.Builder()
+            .connectTimeout(30, TimeUnit.SECONDS)
+            .readTimeout(30, TimeUnit.SECONDS)
```
CandelR: I think adding a [writeTimeout](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/-builder/write-timeout.html) could also save us future problems when uploading files. Also is gonna be the same for all I think we can use [callTimeout](https://square.github.io/okhttp/5.x/okhttp/okhttp3/-ok-http-client/-builder/call-timeout.html?query=fun%20callTimeout(duration:%20Duration):%20OkHttpClient.Builder). What do you think?

android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
terrerox: Ive changed it in this pr https://github.com/internxt/drive-mobile/pull/440/changes

terrerox:
@CandelR this one is ready for your review

CandelR (approved):

PR #423, opened 2026-04-20
[PB-6093, 6094, 6095] feature/Implemented enable photos logic and persist related states

CandelR:
- Implemented enable photos logic and persist related states


src/hooks/common/useUseCase/useUseCase.spec.tsx
```diff
@@ -1,15 +1,14 @@
 import { renderHook, waitFor } from '@testing-library/react-native';
 import { useUseCase } from './useUseCase';
 
-describe('useUseCase hook', () => {
-  test('Should resolve with data once the use case function is completed without errors', async () => {
+describe('useUseCase', () => {
+  it('when use case resolves successfully, then data is set and no error', async () => {
```
larryrider: I think this should be changed again to "test" instead of "it"

src/services/photos/photoPermissionService.spec.ts
```diff
@@ -0,0 +1,118 @@
...
+  ...overrides,
+});
+
+describe('PhotoPermissionService', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('getStatus', () => {
+    it('when iOS accessPrivileges is all, then returns granted', async () => {
```
larryrider: same with all this tests.
The "it" tests should have this type of texts: "it should ... when ...." or "it does ... when ...."

src/store/slices/photos/index.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+
+const getPersistedState = (): PhotosState =>
+  JSON.parse((mockAsyncStorage.saveItem.mock.calls.at(-1) as [string, string])[1]);
+
+describe('photos slice', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('when store initializes, then backup is disabled with wifi-only and undetermined permission', () => {
```
larryrider: same here

larryrider (approved):

PR #430, opened 2026-04-23
[PB-5926]: show Internxt Drive in SAF picker per session state

terrerox:
InternxtDocumentsProvider.queryRoots() now reads credentials from EncryptedSharedPreferences via a new InternxtAuthManager and returns an empty cursor when no user is signed in, so "Internxt Drive" only appears in the system file picker while the app has an active session; credentials are written by a new RN bridge (InternxtAuthCredentialsModule) from signInThunk and refreshTokensThunk and cleared from signOutThunk, with both paths firing contentResolver.notifyChange(rootsUri) so the picker refreshes without restarting the app, while queryRoots() itself stays strictly synchronous and local-only to avoid freezing the system picker.

### DIFF
--- a/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
@@ -4,22 +4,35 @@ import android.database.Cursor
 import android.database.MatrixCursor
 import android.os.CancellationSignal
 import android.os.ParcelFileDescriptor
+import android.provider.DocumentsContract
 import android.provider.DocumentsContract.Document
 import android.provider.DocumentsContract.Root
 import android.provider.DocumentsProvider
 import com.internxt.cloud.R
+import com.internxt.cloud.documents.auth.InternxtAuthManager
 
 class InternxtDocumentsProvider : DocumentsProvider() {
 
-    override fun onCreate(): Boolean = true
+    private lateinit var authManager: InternxtAuthManager
+
+    override fun onCreate(): Boolean {
+        authManager = InternxtAuthManager.create(context!!.applicationContext)
+        return true
+    }
 
     override fun queryRoots(projection: Array<String>?): Cursor {
         val cursor = MatrixCursor(resolveRootProjection(projection))
+        val ctx = context ?: return cursor
+        cursor.setNotificationUri(ctx.contentResolver, DocumentsContract.buildRootsUri(AUTHORITY))
+
+        val rootUuid = authManager.authenticatedRootUuid() ?: return cursor
+
         cursor.newRow().apply {
             add(Root.COLUMN_ROOT_ID, ROOT_ID)
-            add(Root.COLUMN_DOCUMENT_ID, ROOT_DOCUMENT_ID)
-            add(Root.COLUMN_TITLE, context?.getString(R.string.documents_provider_label))
-            add(Root.COLUMN_FLAGS, 0)
+            add(Root.COLUMN_DOCUMENT_ID, rootUuid)
+            add(Root.COLUMN_TITLE, ctx.getString(R.string.documents_provider_label))
+            authManager.userEmail()?.let { add(Root.COLUMN_SUMMARY, it) }
+            add(Root.COLUMN_FLAGS, Root.FLAG_SUPPORTS_CREATE or Root.FLAG_SUPPORTS_IS_CHILD)
             add(Root.COLUMN_ICON, R.mipmap.ic_launcher)
         }
         return cursor
@@ -50,8 +63,7 @@ class InternxtDocumentsProvider : DocumentsProvider() {
 
     companion object {
         const val AUTHORITY = "com.internxt.cloud.documents"
-        private const val ROOT_ID = "root"
-        private const val ROOT_DOCUMENT_ID = "root"
+        private const val ROOT_ID = "internxt-root"
 
         private val DEFAULT_ROOT_PROJECTION = arrayOf(
             Root.COLUMN_ROOT_ID,
--- a/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
@@ -16,7 +16,8 @@ class InternxtDocumentsProvider : DocumentsProvider() {
     private lateinit var authManager: InternxtAuthManager
 
     override fun onCreate(): Boolean {
-        authManager = InternxtAuthManager.create(context!!.applicationContext)
+        val ctx = context ?: return false
+        authManager = InternxtAuthManager.create(ctx.applicationContext) ?: return false
         return true
     }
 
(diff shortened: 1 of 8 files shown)


android/app/src/main/java/com/internxt/cloud/documents/auth/InternxtAuthManager.kt
```diff
@@ -0,0 +1,101 @@
...
+        prefs.edit()
+            .putString(KEY_BEARER_TOKEN, creds.bearerToken)
+            .putString(KEY_USER_ID, creds.userId)
+            .putString(KEY_BRIDGE_USER, creds.bridgeUser)
+            .putString(KEY_ROOT_FOLDER_UUID, creds.rootFolderUuid)
+            .putString(KEY_EMAIL, creds.email)
+            .putString(KEY_DRIVE_BASE_URL, creds.driveBaseUrl)
+            .putString(KEY_BRIDGE_BASE_URL, creds.bridgeBaseUrl)
+            .putString(KEY_DESKTOP_TOKEN, creds.desktopToken)
+            .apply()
```
CandelR: apply is async, maybe for this it is preferable use commit

src/store/slices/auth/index.ts
```diff
@@ -33,6 +35,22 @@ const initialState: AuthState = {
...
+async function syncNativeCredentials(token: string, user: UserSettings): Promise<void> {
+  try {
+    await setCredentials({
+      bearerToken: token,
+      userId: user.userId,
+      bridgeUser: user.bridgeUser,
+      rootFolderUuid: user.rootFolderUuid || user.rootFolderId,
+      email: user.email,
+      driveBaseUrl: appService.constants.DRIVE_NEW_API_URL,
+      bridgeBaseUrl: appService.constants.BRIDGE_URL,
```
CandelR, line 57: The desktopToken is missing

src/store/slices/auth/index.ts
```diff
@@ -33,6 +35,22 @@ const initialState: AuthState = {
   sessionPassword: undefined,
 };
 
+async function syncNativeCredentials(token: string, user: UserSettings): Promise<void> {
+  try {
+    await setCredentials({
+      bearerToken: token,
+      userId: user.userId,
+      bridgeUser: user.bridgeUser,
+      rootFolderUuid: user.rootFolderUuid || user.rootFolderId,
```
CandelR: `rootFolderId` as fallback is wrong, remember that `rootFolderId` is the numeric remote database id and the rootFolderUuid is the uuid one. For api v2 we use `rootFolderUuid` :)

android/app/src/main/java/com/internxt/cloud/documents/auth/InternxtAuthManager.kt
```diff
@@ -0,0 +1,101 @@
+package com.internxt.cloud.documents.auth
+
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.security.crypto.EncryptedSharedPreferences
+import androidx.security.crypto.MasterKey
+import com.internxt.cloud.BuildConfig
+import com.internxt.cloud.documents.api.AuthConfig
+
+class InternxtAuthManager(private val prefs: SharedPreferences) {
```
CandelR, line 13: Why is this AuthManager in docuemtns instead of the other auth directory?

android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
```diff
@@ -4,22 +4,35 @@ import android.database.Cursor
...
 import com.internxt.cloud.R
+import com.internxt.cloud.documents.auth.InternxtAuthManager
 
 class InternxtDocumentsProvider : DocumentsProvider() {
 
-    override fun onCreate(): Boolean = true
+    private lateinit var authManager: InternxtAuthManager
+
+    override fun onCreate(): Boolean {
+        authManager = InternxtAuthManager.create(context!!.applicationContext)
```
CandelR: Be carefull with not null assertion operator, I think this logic could lead in a unexpected bugs, better  return false if there are no context
```
val ctx = context ?: return false
authManager = InternxtAuthManager.create(ctx.applicationContext)
return true
```

android/app/src/main/java/com/internxt/cloud/documents/auth/InternxtAuthManager.kt
```diff
@@ -0,0 +1,101 @@
...
+                .build()
+            val prefs = EncryptedSharedPreferences.create(
+                context,
+                PREFS_FILE,
+                masterKey,
+                EncryptedSharedPreferences.PrefKeyEncryptionScheme.AES256_SIV,
+                EncryptedSharedPreferences.PrefValueEncryptionScheme.AES256_GCM,
+            )
+            return InternxtAuthManager(prefs)
+        }
```
CandelR: I think that is recommendable to wrap this logic in a try/catch and return null on failure so the provider degrades gracefully instead of crashing      
                                                                                                                                   
```
fun create(context: Context): InternxtAuthManager? {
      return try {     
          ...
      } catch (e: GeneralSecurityException) {
          Log.e(TAG, "Keystore unavailable, SAF auth disabled", e)                                                                   
          null
      } catch (e: IOException) {                                                                                                     
          Log.e(TAG, "Could not open encrypted prefs, SAF auth disabled", e)                                                       
          null                                                                                                                       
      }                
  }                      
```                                                                                                            
If implement something like the example, then we need to handle the nullable return in `onCreate()`.
We could manage as if `create()` returns null, `onCreate` return false

android/app/src/main/java/com/internxt/cloud/documents/auth/InternxtAuthManager.kt
terrerox: Because only the file picker reads these credentials. The auth/ folder is just the JS <-> Android bridge, it writes, but doesn't own the data.

android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
```diff
@@ -4,22 +4,36 @@ import android.database.Cursor
...
+        authManager = InternxtAuthManager.create(ctx.applicationContext) ?: return false
+        return true
+    }
 
     override fun queryRoots(projection: Array<String>?): Cursor {
         val cursor = MatrixCursor(resolveRootProjection(projection))
+        val ctx = context ?: return cursor
+        cursor.setNotificationUri(ctx.contentResolver, DocumentsContract.buildRootsUri(AUTHORITY))
+
+        val rootUuid = authManager.authenticatedRootUuid() ?: return cursor
```
CandelR: I’ve just realised that `authManager` shouldn’t be `lateinit`. 
If `InternxtAuthManager.create()` fails and `onCreate()` returns `false`, the variable remains uninitialised.
In `queryRoots()`, `authManager.authenticatedRootUuid()` is called directly without a safe call, so if Android calls `queryRoots()` in that state, the app throws an `UninitializedPropertyAccessException` in the provider process without any visible crash.
I think that we need to remove the late init from authManager and initialise it `null`
`private var authManager: InternxtAuthManager? = null` and using a safe call in `queryRoots()`:
I think that with this latest change, this PR should now be ready :)

CandelR (approved):

PR #432, opened 2026-04-27
[PB-6089,PB-6090] feature/Initial gallery scan

CandelR:
- Implemented photo backup discovery process (database, redux actions and state and UI

### DIFF
--- a/src/services/photos/database/tables/asset_sync.ts
+++ b/src/services/photos/database/tables/asset_sync.ts
@@ -1,0 +1,56 @@
+const TABLE_NAME = 'asset_sync';
+
+const statements = {
+  createTable: `
+    CREATE TABLE IF NOT EXISTS ${TABLE_NAME} (
+      asset_id        TEXT    PRIMARY KEY NOT NULL,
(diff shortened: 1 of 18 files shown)


larryrider (approved):

src/services/photos/database/tables/asset_sync.ts
```diff
@@ -0,0 +1,56 @@
...
+    );
+  `,
+  createIndex: `CREATE INDEX IF NOT EXISTS idx_asset_sync_status ON ${TABLE_NAME}(status);`,
+  dropTable: `DROP TABLE IF EXISTS ${TABLE_NAME};`,
+
+  markPending: `
+    INSERT INTO ${TABLE_NAME} (asset_id, status)
+    VALUES (?, 'pending')
+    ON CONFLICT(asset_id) DO UPDATE SET status = 'pending'
+    WHERE ${TABLE_NAME}.status != 'synced';
```
larryrider, line 23: if an asset is marked as "synced", then it will never be marked again as "error" or "pending", is that right?

CandelR: Correct! :D

PR #434, opened 2026-04-29
feat(android): list folder contents in SAF documents provider

terrerox:
- Implement queryChildDocuments to paginate folders + files for a parent UUID
  - Implement queryDocument to resolve real metadata via folders/{uuid}/meta and files/{uuid}/meta, with a UUID-named placeholder fallback for offline / transient failures
  - Add InternxtApiClient.getFolder / getFile (404 → null) plus tests
  - Introduce DocumentRowBuilder (column-keyed rows) and MimeTypes (hand-rolled table + Android MimeTypeMap fallback) with tests
  - Surface InternxtApiException + null auth-config via Log.w so silent failures are no longer invisible
  - Fix auth slice to sync newToken (not token) to native credentials so the documents provider authenticates against drive endpoints

### DIFF
--- a/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
@@ -8,7 +8,10 @@ import android.provider.DocumentsContract
 import android.provider.DocumentsContract.Document
 import android.provider.DocumentsContract.Root
 import android.provider.DocumentsProvider
+import android.util.Log
 import com.internxt.cloud.R
+import com.internxt.cloud.documents.api.InternxtApiClient
+import com.internxt.cloud.documents.api.InternxtApiException
 import com.internxt.cloud.documents.auth.InternxtAuthManager
 
 class InternxtDocumentsProvider : DocumentsProvider() {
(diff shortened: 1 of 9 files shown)


android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
```diff
@@ -29,6 +29,16 @@ class InternxtApiClient(
     fun listFolderFiles(parentUuid: String, offset: Int = 0, limit: Int = DEFAULT_PAGE_SIZE): List<DriveFile> =
         listChildren(parentUuid, kind = "files", jsonKey = "files", offset, limit, ::parseFile)
 
+    fun getFolder(uuid: String): DriveFolder? = getMetaOrNull("folders/$uuid/meta", ::parseFolder)
+
+    fun getFile(uuid: String): DriveFile? = getMetaOrNull("files/$uuid/meta", ::parseFile)
+
+    private fun <T> getMetaOrNull(path: String, parse: (JSONObject) -> T): T? = try {
+        parse(executeApiRequest(driveRequest(driveUrl(path)).get().build()))
+    } catch (_: InternxtApiException.NotFoundException) {
```
CandelR, line 38: What happens here if exceptions other than `NotFoundException` are thrown?

android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
```diff
@@ -38,14 +41,89 @@ class InternxtDocumentsProvider : DocumentsProvider() {
...
+        return cursor
+    }
 
     override fun queryChildDocuments(
         parentDocumentId: String?,
         projection: Array<String>?,
         sortOrder: String?
-    ): Cursor = MatrixCursor(resolveDocumentProjection(projection))
+    ): Cursor {
+        val cursor = MatrixCursor(resolveDocumentProjection(projection))
```
CandelR, line 104: I would say that the current implementation blocks the provider thread until all pages have been fetched, so the user sees nothing until the entire folder has been loaded. This works, but it isn’t scalable for large folders. The correct approach is to return the first page immediately, abd tgeb load the remaining pages in a background coroutine

android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
```diff
@@ -38,14 +41,89 @@ class InternxtDocumentsProvider : DocumentsProvider() {
...
+            cursor.addDocumentRow(
+                DocumentRowBuilder.folderRow(id, ctx.getString(R.string.documents_provider_label))
+            )
+            return cursor
+        }
+
+        val api = apiClient(op = "queryDocument")
+        val row = if (api == null) null else try {
+            api.getFolder(id)?.let { DocumentRowBuilder.folderRow(it) }
+                ?: api.getFile(id)?.let { DocumentRowBuilder.fileRow(it) }
```
CandelR: I think that this logic makes an unnecessary call, it would almost be better to add a way of distinguishing between folders and files, as this code first checks whether it’s a folder and, if not, tries with a file, if I’m not mistaken.

android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
terrerox: Unauthorized, ApiError, NetworkException, or a JSONException from a malformed response, is passed back to the caller. That keeps “missing” separate from actual failures, so the provider can log, retry, or report
`EXTRA_ERROR` instead of quietly treating network or auth problems as “not found.”

CandelR (commented):
The app isn't compiling because of the Kotlin part in this branch. Have you sorted it out in another branch?

terrerox:
> The app isn't compiling because of the Kotlin part in this branch. Have you sorted it out in another branch?

Yes! I fixed it in this [PR](https://github.com/internxt/drive-mobile/pull/440)

CandelR (approved):

PR #436, opened 2026-04-30
[PB-6085] feature/Added photo backup upload service layer

CandelR:
Introduces the core services for photo backup: asset scanning, deduplication via local SQLite, folder hierarchy management, and encrypted upload to the network bucket. Assets are tracked by sync status (pending, pending_edit, synced, error) with modificationTime to detect edits after initial sync.    

  ## Summary                                                                                                                         
  - Adds `PhotoAssetScanner` to paginate device media library assets
  - Adds `PhotoDeduplicator` to diff scanned assets against local DB (new vs edited)                                                 
  - Adds `PhotoUploadService` with iOS/Android URI resolution, encryption, and bucket upload                                         
  - Adds `PhotoUploadQueue` for concurrent uploads                                              
  - Adds `PhotoBackupFolders` to lazily create and cache the remote folder hierarchy (device/year/month/day)                         
  - Extends `photosLocalDB` with `pending_edit` status and `modificationTime` tracking

### DIFF
--- a/src/services/photos/PhotoAssetScanner.ts
+++ b/src/services/photos/PhotoAssetScanner.ts
@@ -1,21 +1,44 @@
 import * as MediaLibrary from 'expo-media-library';
 import { logger } from 'src/services/common';
 
+const PAGE_SIZE = 200;
+const MEDIA_TYPES = [MediaLibrary.MediaType.photo, MediaLibrary.MediaType.video];
+
+// onAssetFetched returns true to stop pagination early
+const paginateAssets = async (
+  options: Omit<MediaLibrary.AssetsOptions, 'first' | 'after'>,
+  onAssetFetched: (asset: MediaLibrary.Asset) => boolean,
+): Promise<void> => {
+  let cursor: string | undefined;
+  do {
+    const page = await MediaLibrary.getAssetsAsync({ first: PAGE_SIZE, after: cursor, ...options });
+    for (const asset of page.assets) {
+      if (onAssetFetched(asset)) return;
+    }
+    cursor = page.hasNextPage ? page.endCursor : undefined;
+  } while (cursor);
+};
+
 export const PhotoAssetScanner = {
+  async getAssetsByIds(assetIds: string[]): Promise<MediaLibrary.Asset[]> {
+    const pendingIdSet = new Set(assetIds);
+    const resolvedAssets: MediaLibrary.Asset[] = [];
+
+    await paginateAssets({ mediaType: MEDIA_TYPES }, (asset) => {
+      if (pendingIdSet.has(asset.id)) resolvedAssets.push(asset);
+      return resolvedAssets.length === assetIds.length;
+    });
+
+    return resolvedAssets;
+  },
+
   async scanAll(): Promise<MediaLibrary.Asset[]> {
     const results: MediaLibrary.Asset[] = [];
-    let cursor: string | undefined;
 
-    do {
-      const batch = await MediaLibrary.getAssetsAsync({
-        first: 200,
-        after: cursor,
-        mediaType: [MediaLibrary.MediaType.photo, MediaLibrary.MediaType.video],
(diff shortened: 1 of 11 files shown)


src/services/photos/database/tables/asset_sync.ts
```diff
@@ -24,6 +24,13 @@ const statements = {
     WHERE ${TABLE_NAME}.status != 'synced';
   `,
 
+  markPendingEdit: `
+    INSERT INTO ${TABLE_NAME} (asset_id, status)
+    VALUES (?, 'pending_edit')
+    ON CONFLICT(asset_id) DO UPDATE SET status = 'pending_edit'
```
sg-gs, line 30: what does `pending_edit` mean?

src/services/photos/PhotoAssetScanner.ts
```diff
@@ -1,21 +1,43 @@
...
+
+const paginateAssets = async (
+  options: Omit<MediaLibrary.AssetsOptions, 'first' | 'after'>,
+  onAssetFetched: (asset: MediaLibrary.Asset) => boolean,
+): Promise<void> => {
+  let cursor: string | undefined;
+  do {
+    const page = await MediaLibrary.getAssetsAsync({ first: PAGE_SIZE, after: cursor, ...options });
+    for (const asset of page.assets) {
+      if (onAssetFetched(asset)) return;
```
sg-gs: Why?

src/services/photos/PhotoAssetScanner.ts
```diff
@@ -1,21 +1,43 @@
...
-        mediaType: [MediaLibrary.MediaType.photo, MediaLibrary.MediaType.video],
-        sortBy: MediaLibrary.SortBy.modificationTime,
-      });
-      results.push(...batch.assets);
-      cursor = batch.hasNextPage ? batch.endCursor : undefined;
-    } while (cursor);
+
+    await paginateAssets({ mediaType: MEDIA_TYPES, sortBy: MediaLibrary.SortBy.modificationTime }, (asset) => {
+      results.push(asset);
+      return false;
```
sg-gs: Why?

src/services/photos/database/tables/asset_sync.ts
CandelR: Is used to mark assets that have been edited but have yet to be uploaded

src/services/photos/PhotoAssetScanner.ts
CandelR: It’s an early return. Initially, the `scanAll` and `getAssetsByIds` functions were separate, when asset retrieval was consolidated using the `MediaLibrary` library, this was added to stop the process. Perhaps I should separate them again, as having `onAssetFetched` return a boolean to indicate whether to stop paginating assets or not is a bit confusing

src/services/photos/PhotoAssetScanner.ts
CandelR: So that it never stops fetching assets, as we want to fetch them all. In any case, as I mentioned in my previous comment, perhaps the best approach is to refactor the common funciton so that the only thing the functions   `scanAll` and `getAssetsByIds` have in common is that they use `MediaLibrary.getAssetsAsync`

src/services/photos/PhotoAssetScanner.ts
sg-gs: Yes, this code is a bit weird to understand

src/services/photos/PhotoAssetScanner.ts
sg-gs: I do not know if the common function has something to see here or not, but I understand that stopping a loop should be obvious and these booleans are a bit random as they do not suggest anything themselves. Returning `false` on the callback of an async function called `paginateAssets` does not suggest that, could you make it clearer?

larryrider (approved):
lgtm, just check sonarcloud issues before merge

PR #437, opened 2026-04-30
[PB-6085] feature/Added unit tests for photo upload service layer

CandelR:
Covers PhotoBackupFolders caching behaviour, PhotoDeduplicator new/edited asset detection, PhotoDeviceId getOrCreate logic, PhotoUploadQueue success/error/replace flows, and PhotoUploadService URI parsing utilities.                                        
                                                                                                                                                                                                               
  ## Summary                                                                                                                         
  - Unit tests for all services introduced in [Photo backup upload services PR](https://github.com/internxt/drive-mobile/pull/436)
  - Tests covering folder caching, deduplication, device ID persistence, upload queue concurrency, and URI parsing edge cases

### DIFF
--- a/src/services/photos/PhotoBackupFolders.spec.ts
+++ b/src/services/photos/PhotoBackupFolders.spec.ts
@@ -57,6 +57,7 @@ describe('PhotoBackupFolders.getOrCreateFolderForDate', () => {
     const folderNames = mockCreateFolder.mock.calls.map((call) => call[1]);
     const monthCalls = folderNames.filter((name) => name === '06');
     expect(monthCalls).toHaveLength(1);
+    expect(mockCreateFolder).toHaveBeenCalledTimes(6);
   });
 
   test('when called for two dates in different years, then separate year folders are created', async () => {
--- a/src/services/photos/PhotoBackupFolders.spec.ts
+++ b/src/services/photos/PhotoBackupFolders.spec.ts
@@ -1,0 +1,97 @@
+import { createFolderWithMerge } from 'src/services/drive/folder/folderOrchestration.service';
+import { photoBackupFolders } from './PhotoBackupFolders';
+
+jest.mock('src/services/AsyncStorageService', () => ({
+  __esModule: true,
+  default: {
+    getUser: jest.fn().mockResolvedValue({ rootFolderId: 'root-uuid' }),
+  },
+}));
+
+jest.mock('src/services/drive/folder/folderOrchestration.service', () => ({
+  createFolderWithMerge: jest.fn(),
+}));
+
+const mockCreateFolder = createFolderWithMerge as jest.Mock;
+
+const DEVICE_ID = 'device-123';
+
+beforeEach(() => {
+  jest.clearAllMocks();
+  photoBackupFolders.clearCache();
+  mockCreateFolder
+    .mockResolvedValueOnce('photos-root-uuid')
+    .mockResolvedValueOnce('device-uuid')
+    .mockResolvedValueOnce('year-uuid')
+    .mockResolvedValueOnce('month-uuid')
+    .mockResolvedValueOnce('day-uuid');
+});
+
+describe('PhotoBackupFolders.getOrCreateFolderForDate', () => {
+  test('when called for a date, then creates the full folder hierarchy and returns the day folder uuid', async () => {
+    const uuid = await photoBackupFolders.getOrCreateFolderForDate(DEVICE_ID, new Date('2024-06-15'));
+    expect(uuid).toBe('day-uuid');
+    expect(mockCreateFolder).toHaveBeenCalledTimes(5);
+  });
+
+  test('when called twice for the same date, then creates folders only once', async () => {
+    await photoBackupFolders.getOrCreateFolderForDate(DEVICE_ID, new Date('2024-06-15'));
+    await photoBackupFolders.getOrCreateFolderForDate(DEVICE_ID, new Date('2024-06-15'));
+
+    expect(mockCreateFolder).toHaveBeenCalledTimes(5);
+  });
+
+  test('when called for two dates in the same month, then the month folder is created only once', async () => {
+    mockCreateFolder
+      .mockReset()
+      .mockResolvedValueOnce('photos-root-uuid')
+      .mockResolvedValueOnce('device-uuid')
+      .mockResolvedValueOnce('year-uuid')
+      .mockResolvedValueOnce('month-uuid')
+      .mockResolvedValueOnce('day-15-uuid')
+      .mockResolvedValueOnce('day-20-uuid');
+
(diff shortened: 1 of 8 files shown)


src/services/photos/PhotoBackupFolders.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+      .mockResolvedValueOnce('month-uuid')
+      .mockResolvedValueOnce('day-15-uuid')
+      .mockResolvedValueOnce('day-20-uuid');
+
+    await photoBackupFolders.getOrCreateFolderForDate(DEVICE_ID, new Date('2024-06-15'));
+    await photoBackupFolders.getOrCreateFolderForDate(DEVICE_ID, new Date('2024-06-20'));
+
+    const folderNames = mockCreateFolder.mock.calls.map((call) => call[1]);
+    const monthCalls = folderNames.filter((name) => name === '06');
+    expect(monthCalls).toHaveLength(1);
```
TamaraFinogina, line 59: shouldn't it also be checked expect(mockCreateFolder).toHaveBeenCalledTimes(9); (5 for May 15 and 4 for May 20)?

src/services/photos/PhotoBackupFolders.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+  mockCreateFolder
+    .mockResolvedValueOnce('photos-root-uuid')
+    .mockResolvedValueOnce('device-uuid')
+    .mockResolvedValueOnce('year-uuid')
+    .mockResolvedValueOnce('month-uuid')
+    .mockResolvedValueOnce('day-uuid');
+});
+
+describe('PhotoBackupFolders.getOrCreateFolderForDate', () => {
+  test('when called for a date, then creates the full folder hierarchy and returns the day folder uuid', async () => {
```
TamaraFinogina, line 31: Should there be a test for a date in the future? something like: does not create anything for the future date?

src/services/photos/PhotoDeviceId.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+    mockSetItem.mockResolvedValue(undefined);
+
+    const id = await PhotoDeviceId.getOrCreate();
+
+    expect(id).toBe('generated-uuid');
+    expect(mockSetItem).toHaveBeenCalledTimes(1);
+    expect(mockSetItem).toHaveBeenCalledWith(expect.any(String), 'generated-uuid');
+  });
+
+  test('when no device ID is stored, then successive calls generate only one ID per call', async () => {
```
TamaraFinogina, line 42: And if a stored device ID is different? Should we have a test for that too? Say the user has 2 different phones and 1 account

src/services/photos/PhotoUploadQueue.spec.ts
```diff
@@ -0,0 +1,107 @@
+import * as MediaLibrary from 'expo-media-library';
+import { AssetUploadJob, PhotoUploadQueue } from './PhotoUploadQueue';
+import { PhotoUploadService } from './PhotoUploadService';
+
+jest.mock('./PhotoUploadService', () => ({
```
TamaraFinogina, line 5: btw, maybe with Vitest directly it will be easier

src/services/photos/PhotoUploadService.utils.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+  });
+});
+
+describe('stripFileScheme', () => {
+  test('when a file URI is provided, then it returns the decoded path without the scheme', () => {
+    expect(stripFileScheme('file:///data/user/0/com.app/photo.jpg')).toBe('/data/user/0/com.app/photo.jpg');
+  });
+
+  test('when the path contains percent-encoded characters, then it decodes them', () => {
+    expect(stripFileScheme('file:///data/My%20Camera/photo.jpg')).toBe('/data/My Camera/photo.jpg');
```
TamaraFinogina, line 34: will it work for extension too?

 expect(stripFileScheme('file:///downloads/photo.jpg%2Eexe')).toBe('/downloads/photo.jpg.exe');

src/services/photos/PhotoUploadService.utils.spec.ts
```diff
@@ -0,0 +1,79 @@
...
+});
+
+describe('splitFileNameAndExtension', () => {
+  test('when a filename with an extension is provided, then it splits correctly', () => {
+    expect(splitFileNameAndExtension('photo.jpg')).toEqual({ plainName: 'photo', fileExtension: 'jpg' });
+  });
+
+  test('when a filename has multiple dots, then it uses the last dot as the separator', () => {
+    expect(splitFileNameAndExtension('my.vacation.photo.jpg')).toEqual({
+      plainName: 'my.vacation.photo',
```
TamaraFinogina, line 71: will it do this too? Cause maybe we'll need to blacklist some extensions

expect(splitFileNameAndExtension('my.vacation.photo.apk')).toEqual({
    plainName: 'my.vacation.photo',
    fileExtension: 'apk',
  });

src/services/photos/PhotoBackupFolders.spec.ts
CandelR: The folder is created using the asset's metadata, so if it indicates a future date, it will be created based on that. It may be that the device's time is incorrect, and we shouldn't block the copying of any assets

src/services/photos/PhotoDeviceId.spec.ts
CandelR: then it will have 2 ids, the id is for device

src/services/photos/PhotoUploadQueue.spec.ts
CandelR: I’d say that Vitests is specifically designed for Vite, so it’s not the right choice for a React Native project

src/services/photos/PhotoUploadService.utils.spec.ts
CandelR: This feature only scans media assets. Basically, what you can see in a gallery app

src/services/photos/PhotoBackupFolders.spec.ts
CandelR: We can add it, but there will be `expect(mockCreateFolder).toHaveBeenCalledTimes(6)`, not 9. There’s a cache to prevent unnecessary calls :)

src/services/photos/PhotoUploadService.utils.spec.ts
CandelR: Yes, it works, and that test could be added. However, this is only used for media assets, with all that that entails in terms of file extensions

TamaraFinogina (approved):
Sonar complains about duplicated code

CandelR:
> Sonar complains about duplicated code

Yeah, it’s because of the tests. I’ve tried to stop it from checking the duplicate test code, but Sonar just ignores it :(

PR #440, opened 2026-04-30
feat(android): support rename, move, and trash in SAF picker

terrerox:
- Override renameDocument, moveDocument, and deleteDocument in InternxtDocumentsProvider so SAF surfaces the corresponding row actions.
  - Surface FLAG_SUPPORTS_RENAME | FLAG_SUPPORTS_DELETE | FLAG_SUPPORTS_MOVE on file and folder rows; keep root and placeholder rows mutation-free.
  - Cache item kind (file vs folder) per UUID so mutations resolve their endpoint without an extra round-trip on the happy path.
  - Notify affected child-document URIs after each mutation so the picker refreshes without manual reload.
  - Fix InternxtApiClient.renameFile / renameFolder to match the SDK: PUT /{files|folders}/{uuid}/meta with { plainName }, void response.
  - Await clearCredentials in signOutThunk so the SAF root is hidden immediately after logout instead of lingering until documentsui refreshes.

### DIFF
--- a/android/app/src/main/java/com/internxt/cloud/documents/DocumentRowBuilder.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/DocumentRowBuilder.kt
@@ -8,13 +8,22 @@ import java.time.format.DateTimeParseException
 
 object DocumentRowBuilder {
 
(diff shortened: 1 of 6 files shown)


CandelR (approved):
Working fine, good job @terrerox 🚀

PR #445, opened 2026-05-08
[PB-6372] featur/Pull to refresh feature

CandelR:
Implemented pull to refresh feature

## Summary
  - Adds pull-to-refresh gesture to the Photos timeline
  - On pull: reloads local gallery from MediaLibrary and triggers a forced cloud history sync (TTL bypassed) in parallel
  - Cloud-only photos appear progressively, no data is deleted from the local DB, existing records are upserted
  - Guard prevents concurrent refreshes: skips if backup is disabled, a scan/upload is in progress, or a cloud sync is already running

### DIFF
--- a/src/screens/PhotosScreen/index.tsx
+++ b/src/screens/PhotosScreen/index.tsx
@@ -5,7 +5,10 @@ import AppScreen from 'src/components/AppScreen';
 import { useAppDispatch, useAppSelector } from 'src/store/hooks';
 import { forceRefreshThunk, photosActions, runBackupCycleThunk } from 'src/store/slices/photos';
 import { useTailwind } from 'tailwind-rn';
+import strings from '../../../assets/lang/strings';
 import { photoPermissionService } from '../../services/photos/photoPermissionService';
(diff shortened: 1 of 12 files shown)


src/screens/PhotosScreen/index.tsx
```diff
@@ -29,6 +30,15 @@ const PhotosScreen = (): JSX.Element => {
   const listHeader =
     accessState.type === 'backup-off' ? <BackupDisabledBanner onEnablePress={handleEnableBackup} /> : undefined;
 
+  const handleRefresh = useCallback(async () => {
+    setRefreshing(true);
+    try {
+      await Promise.all([reloadLocal(), dispatch(forceRefreshThunk()).unwrap()]);
```
larryrider: If one of them fails, should the other one fail too?

If not, then its better to use Promise.allSettled 🤔

CandelR: Yeah, absolutely. I'll add some notifications as well, just in case they fail notify the user correctly

larryrider (approved):

PR #451, opened 2026-05-12
[PB-5928]: Download & decryption in the SAF documents provider

terrerox:
## PB-5928 — Download & decryption in the SAF documents provider

  Adds end-to-end read-only file access to `InternxtDocumentsProvider`: `openDocument` now fetches encrypted shards from the bridge, decrypts them
  locally with the user's mnemonic, caches the result, and hands SAF a read-only `ParcelFileDescriptor`. Previously this method threw
  `UnsupportedOperationException`.

  ## Summary

  - **Open path (`InternxtDocumentsProvider.openDocument`)**: runs on a dedicated executor, decoded into `requireFileMetadata → materializeIntoCache
   → openCached`. Handles `CancellationSignal`, falls back to a hot cache on `${uuid}_${updatedAt-slug}`, and cleans up temp files on cancel/error.
  - **`EncryptedFileDownloader`**: streams shards in `index` order into a single output file via OkHttp, wires `CancellationSignal` to cancel the
  active call, and surfaces cancellations as `OperationCanceledException`.
  - **`FileKeyDeriver`**: derives the AES-CTR key/IV from `(mnemonic, bucketId, index)` using `CryptoService` — matches the JS
  `src/network/crypto.ts: generateFileKey` algorithm (PBKDF2 → SHA-512 chain). Decryption itself goes through `CryptoService.decryptFile` (native).
  - **`DocumentCache`**: process-local disk cache under `cacheDir/internxt_documents/{cache,tmp}`. Promotes via `rename()`, prunes stale siblings on
   success, wipes temps on close.
  - **Bridge `/info` endpoint**: `getDownloadLinks` now hits `buckets/{bucket}/files/{file}/info` with `x-api-version: 2`, validating `version == 2`
   and non-empty shards.
  - **Auth**: `AuthConfig` and `InternxtAuthManager` now carry the user's `mnemonic` (passed through `InternxtAuthCredentialsModule` from the RN
  auth slice). Required for key derivation; openDocument fails fast with a clear "sign out and back in" message if missing.
  - **`HttpClients`**: single source of truth for OkHttp configuration — short timeouts for API JSON calls, long/no-timeout for shard streaming.
  - **`InternxtApiException.MalformedResponse`**: new sealed variant for client-side response-shape errors (replaces fake `ApiError(500/409/404)`
  thrown by the parser).

### DIFF
--- a/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
@@ -37,6 +37,9 @@ class InternxtDocumentsProvider : DocumentsProvider() {
         @Volatile var errorMessage: String? = null
         val rows = mutableListOf<Map<String, Any?>>()
     }
+    private val itemKinds = ConcurrentHashMap<String, ItemKind>()
+
+    private enum class ItemKind { FILE, FOLDER }
 
     override fun onCreate(): Boolean {
         val ctx = context ?: return false
--- a/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
@@ -5,6 +5,9 @@ import android.database.MatrixCursor
 import android.net.Uri
 import android.os.Bundle
 import android.os.CancellationSignal
+import android.os.Handler
+import android.os.HandlerThread
+import android.os.OperationCanceledException
 import android.os.ParcelFileDescriptor
 import android.provider.DocumentsContract
 import android.provider.DocumentsContract.Document
@@ -16,8 +19,16 @@ import com.internxt.cloud.documents.api.InternxtApiClient
 import com.internxt.cloud.documents.api.InternxtApiException
 import com.internxt.cloud.documents.api.model.TrashItem
 import com.internxt.cloud.documents.auth.InternxtAuthManager
+import com.internxt.cloud.documents.cache.DocumentCache
+import com.internxt.cloud.documents.crypto.FileKeyDeriver
+import com.internxt.cloud.documents.download.EncryptedFileDownloader
+import com.rncrypto.util.CryptoService
+import okhttp3.OkHttpClient
+import java.io.File
 import java.io.FileNotFoundException
 import java.util.concurrent.ConcurrentHashMap
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
 import java.util.concurrent.Executors
 
 class InternxtDocumentsProvider : DocumentsProvider() {
@@ -38,12 +49,22 @@ class InternxtDocumentsProvider : DocumentsProvider() {
         val rows = mutableListOf<Map<String, Any?>>()
     }
     private val itemKinds = ConcurrentHashMap<String, ItemKind>()
+    private lateinit var closeHandler: Handler
+    private val downloadClient: OkHttpClient by lazy {
+        OkHttpClient.Builder()
+            .connectTimeout(30, TimeUnit.SECONDS)
+            .readTimeout(2, TimeUnit.MINUTES)
+            .writeTimeout(2, TimeUnit.MINUTES)
+            .callTimeout(0, TimeUnit.MILLISECONDS)
+            .build()
+    }
 
     private enum class ItemKind { FILE, FOLDER }
 
     override fun onCreate(): Boolean {
         val ctx = context ?: return false
         authManager = InternxtAuthManager.create(ctx.applicationContext) ?: return false
+        closeHandler = Handler(HandlerThread("InternxtDocsClose").apply { start() }.looper)
         return true
     }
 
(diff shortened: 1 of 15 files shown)


sg-gs:
Reduce the kLOC to <500. I will review once it is ready for review @terrerox

android/app/src/main/java/com/internxt/cloud/documents/crypto/FileKeyDeriver.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    private const val PBKDF2_SALT = "mnemonic"
+    private const val PBKDF2_ROUNDS = 2048
+    private const val SEED_LENGTH = 64
+    private const val FILE_KEY_LENGTH = 32
+    private const val IV_LENGTH = 16
+
+    fun deriveFileKey(mnemonic: String, bucketIdHex: String, indexHex: String): ByteArray {
+        val crypto = CryptoService.getInstance()
+        val seed = crypto.pbkdf2(mnemonic, PBKDF2_SALT.toByteArray(Charsets.UTF_8), PBKDF2_ROUNDS, SEED_LENGTH)
```
TamaraFinogina, line 25: Can we explicitly set PBKDF2WithHmacSHA512 to make sure there are no surprises?

android/app/src/main/java/com/internxt/cloud/documents/http/HttpClients.kt
```diff
@@ -0,0 +1,24 @@
...
+            .callTimeout(30, TimeUnit.SECONDS)
+            .build()
+    }
+
+    val download: OkHttpClient by lazy {
+        OkHttpClient.Builder()
+            .connectTimeout(30, TimeUnit.SECONDS)
+            .readTimeout(2, TimeUnit.MINUTES)
+            .writeTimeout(2, TimeUnit.MINUTES)
+            .callTimeout(0, TimeUnit.MILLISECONDS)
```
TamaraFinogina, line 21: why are timeouts different for download and api?

android/app/src/test/java/com/internxt/cloud/documents/api/InternxtApiClientTest.kt
```diff
@@ -148,15 +149,22 @@ class InternxtApiClientTest {
 
         val links = client.getDownloadLinks(BUCKET_ID, "file-id-1")
 
+        assertEquals(BUCKET_ID, links.bucket)
```
TamaraFinogina, line 152: I don't see a test that shows generateFileKey produces the same result as drive-web (i.e., that for the given mnemonic, bucketID, and index, the result generated by generateFileKey here equals the hardcoded string result computed by drive-web). Do we have it?

android/app/src/main/java/com/internxt/cloud/documents/crypto/FileKeyDeriver.kt
terrerox: The `crypto.pdkdf2()` in `@internxt/rn-crypto` already hardcodes the `PBKDF2WithHmacSHA512`

android/app/src/test/java/com/internxt/cloud/documents/api/InternxtApiClientTest.kt
terrerox: No, i added it.

android/app/src/main/java/com/internxt/cloud/documents/http/HttpClients.kt
terrerox: The api client handles small JSON requests (list folders, rename a file). These should take milliseconds, if one takes 30 seconds, something's broken, so we fail fast instead of freezing the UI.

The download client streams files that can be many gigabytes. A 5 GB file on slow WiFi takes way longer than 30 seconds, so applying the same strict timeout would kill every large download mid-stream.

WDYT?

TamaraFinogina (approved):

android/app/src/main/java/com/internxt/cloud/documents/api/InternxtApiClient.kt
```diff
@@ -113,9 +113,37 @@ class InternxtApiClient(
...
+        return parseDownloadLinks(executeApiRequest(request), bucketId, fileId)
+    }
+
+    private fun parseDownloadLinks(body: JSONObject, bucketId: String, fileId: String): DownloadLinks {
+        val index = body.optStringOrNull("index")
+            ?: throw InternxtApiException.MalformedResponse("Bridge /info missing 'index'")
+        val version = body.optInt("version", 1)
+        if (version != 2) {
+            throw InternxtApiException.MalformedResponse("File version=$version is not supported (V2 only)")
+        }
```
CandelR, line 127: Maybe we will need to implement (in another task, no problem) the v1 support. I think that still around 1M files that use that version :)

android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
```diff
@@ -338,7 +359,133 @@ class InternxtDocumentsProvider : DocumentsProvider() {
...
+        val cfg = authManager?.loadAuthConfig() ?: throw FileNotFoundException("Not authenticated")
+        if (cfg.mnemonic.isBlank()) {
+            throw FileNotFoundException("Stored credentials have no mnemonic; sign out and back in")
+        }
+        val api = InternxtApiClient(cfg)
+        val file = requireFileMetadata(api, fileUuid)
+
+        val cacheFile = DocumentCache.cacheFileFor(ctx, id, file.updatedAt)
+        if (cacheFile.exists() && cacheFile.length() > 0) {
+            return openCached(ctx, id, cacheFile)
```
CandelR, line 408: maybe we should check cache before `requireFileMetadata`, i think this wont open cached files if the metadata call fails

TamaraFinogina:
@terrerox Got conflicts now

android/app/src/main/java/com/internxt/cloud/documents/InternxtDocumentsProvider.kt
terrerox: Good catch, but I don't think we should check it before `requireFileMetadata` because it could silently serve a stale file even when the server has a newer version. I implemented a fallback-on-failure approach where, if `requireFileMetadata` fails, it returns the cached version if one exists

CandelR (approved):

PR #455, opened 2026-05-16
[PB-6329]: Enforce max upload file size across upload flows

terrerox:
Adds size validation before upload in single-file, folder, and share extension flows so oversized files are skipped client-side instead of failing mid-upload. Adds a FileSizeExceededModal that surfaces either the skipped file name or the total count, and a storageThunks.ensureMaxUploadFileSizeFresh thunk so the limit reflects the user's current plan before each upload starts.

### DIFF
--- a/assets/lang/strings.ts
+++ b/assets/lang/strings.ts
@@ -3,11 +3,6 @@ import { NotificationType } from '../../src/types';
 import { BiometricAccessType } from '../../src/types/app';
 import { SortType } from '../../src/types/drive/ui';
 
-const fileSizeLimitSuffix = {
-  en: 'the maximum file size allowed by your plan.',
-  es: 'el tamaño máximo permitido por tu plan.',
-};
-
 const translations = {
   en: {
     languages: {
@@ -694,8 +689,8 @@ const translations = {
       },
       FileSizeExceededModal: {
         title: 'File too large',
-        messageWithName: `"{0}" exceeds ${fileSizeLimitSuffix.en}`,
-        messageWithCount: `{0} files exceed ${fileSizeLimitSuffix.en}`,
+        messageWithName: '"{0}" exceeds the maximum file size allowed by your plan.',
+        messageWithCount: '{0} files exceed the maximum file size allowed by your plan.',
       },
       ComingSoonModal: {
         title: 'Coming soon!',
@@ -1558,8 +1553,8 @@ const translations = {
       },
       FileSizeExceededModal: {
         title: 'Archivo demasiado grande',
-        messageWithName: `"{0}" excede ${fileSizeLimitSuffix.es}`,
-        messageWithCount: `{0} archivos exceden ${fileSizeLimitSuffix.es}`,
+        messageWithName: '"{0}" excede el tamaño máximo permitido por tu plan.',
+        messageWithCount: '{0} archivos exceden el tamaño máximo permitido por tu plan.',
       },
       ComingSoonModal: {
         title: '¡Próximamente!',
--- a/assets/lang/strings.ts
+++ b/assets/lang/strings.ts
@@ -355,7 +355,7 @@ const translations = {
         errorNoInternet: 'No internet connection. Please try again.',
         errorSessionExpired: 'Session expired. Please sign in again.',
         errorPrep: 'Could not prepare the file for upload.',
-        errorFileTooLarge: 'The maximum upload size is 300 MB at a time.',
+        errorFileTooLarge: 'File size exceeds the maximum allowed by your plan.',
         errorFileAlreadyExists: 'A file with this name already exists in this folder.',
         errorPaymentRequired: 'Uploading empty files requires a paid plan.',
         uploading: 'Uploading…',
@@ -687,6 +687,10 @@ const translations = {
         message:
           'Uploading empty files is only available for some plans. Please upgrade your plan to use this feature.',
       },
+      FileSizeExceededModal: {
+        title: 'File too large',
+        message: 'File size exceeds the maximum allowed by your plan.',
+      },
       ComingSoonModal: {
         title: 'Coming soon!',
         subtitle: 'Our fantastic devs are working on it, so stay tuned!',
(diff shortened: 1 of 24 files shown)


assets/lang/strings.ts
```diff
@@ -687,6 +692,11 @@ const translations = {
         message:
           'Uploading empty files is only available for some plans. Please upgrade your plan to use this feature.',
       },
+      FileSizeExceededModal: {
+        title: 'File too large',
+        messageWithName: `"{0}" exceeds ${fileSizeLimitSuffix.en}`,
+        messageWithCount: `{0} files exceed ${fileSizeLimitSuffix.en}`,
```
CandelR: The suffix is just to avoid duplication? Add it to the translation directly, no problem. I think is more confusing the current way

src/components/modals/AddModal/hooks/useFolderUpload.ts
```diff
@@ -285,9 +322,24 @@ export const useFolderUpload = ({ uploadAndCreateFileEntry }: { uploadAndCreateF
...
           if (Platform.OS === 'android' && fileNode.uri.startsWith('content://')) {
-            await uploadFolderFileAndroid(fileNode, parentUuid, signal, uploadId, uploadAndCreateFileEntry, handleEmptyFileSkipped);
+            await uploadFolderFileAndroid(
+              fileNode,
+              parentUuid,
+              signal,
+              uploadId,
+              uploadAndCreateFileEntry,
+              handleEmptyFileSkipped,
+              handleFileSizeExceededSkipped,
```
CandelR: I think we can improve the process for checking the maximum size allowed limit. As I understand it, the limit is per file, isn’t it? There’s no limit based on folder size. 
At the start, the size of all the files is obtained, I think it would be better to give a warning before starting the upload, after step `4. Cloud quota check` which checks the remaining space in the account, that might be a good time to warn that there are files exceeding the quota.
What is the expected behaviour when this happens? Just to be clear on the workflow.

src/components/modals/AddModal/index.tsx
```diff
@@ -408,8 +423,17 @@ function AddModal(): JSX.Element {
...
+    const { filesToUpload, filesExcluded } = validateAndFilterFiles(documents, maxUploadFileSize);
+    const [firstExcluded, ...remainingExcluded] = filesExcluded;
+    const hasExcluded = firstExcluded !== undefined;
+    const hasMultipleExcluded = remainingExcluded.length > 0;
+    if (hasExcluded) {
+      const message = hasMultipleExcluded
+        ? strings.formatString(strings.modals.FileSizeExceededModal.messageWithCount, filesExcluded.length)
+        : strings.formatString(strings.modals.FileSizeExceededModal.messageWithName, firstExcluded.name);
+      dispatch(uiActions.setFileSizeExceededMessage(message));
+    }
```
CandelR: I think this would work better in a function that handles the logic for the excluded files and displays the modal. That makes the code snippet easier to maintain

src/shareExtension/hooks/useShareUpload.ts
```diff
@@ -80,6 +88,15 @@ const classifyError = (error: unknown): UploadErrorType => {
   return 'general';
 };
 
+const fetchMaxUploadFileSize = async (): Promise<number> => {
+  try {
+    const res = (await ShareSdkManager.storageV2.getFileVersionLimits()) as { maxUploadFileSize?: number };
```
CandelR, line 93: the return type does not have the field `maxUploadFileSize`? maybe we should add it to the sdk

CandelR (commented):
Looks good @terrerox , just a few things to check ✔️

src/components/modals/AddModal/hooks/useFolderUpload.ts
terrerox: I've improved this process by adding a step 4.1. In the scenario where 1 of multiple files exceeds the limit in a folder, the entire operation is canceled and the modal is shown

src/shareExtension/hooks/useShareUpload.ts
terrerox: When I bumped the SDK version, I noticed a few breaking changes unrelated to this feature, so I split them into this [PR](https://github.com/internxt/drive-mobile/pull/458) to make the review easier

src/components/modals/AddModal/hooks/useFolderUpload.ts
```diff
@@ -65,7 +74,16 @@ const uploadFolderFileAndroid = async (
...
     await StorageAccessFramework.copyAsync({ from: fileNode.uri, to: tempUri });
     if (signal.aborted) return;
-    await uploadAndCreateFileEntry(tempPath, plainName, extension, parentUuid, noopProgress, undefined, undefined, signal);
+    await uploadAndCreateFileEntry(
+      tempPath,
+      plainName,
+      extension,
+      parentUuid,
+      noopProgress,
+      undefined,
```
TamaraFinogina, line 91: I think this was changed to abort signals in sdk PR

TamaraFinogina (approved):

src/components/modals/AddModal/index.tsx
```diff
@@ -418,22 +418,24 @@ function AddModal(): JSX.Element {
...
-  async function uploadDocuments(documents: DocumentPickerFile[]) {
+  const notifyFilesExcludedBySize = (excluded: DocumentPickerFile[]) => {
+    const [firstExcluded, ...remainingExcluded] = excluded;
+    if (firstExcluded === undefined) return;
+    const hasMultipleExcluded = remainingExcluded.length > 0;
+    const message = hasMultipleExcluded
+      ? strings.formatString(strings.modals.FileSizeExceededModal.messageWithCount, excluded.length)
+      : strings.formatString(strings.modals.FileSizeExceededModal.messageWithName, firstExcluded.name);
+    dispatch(uiActions.setFileSizeExceededMessage(message));
+  };
```
CandelR: This function is duplicated, perhaps it would be better to extract it so it can be reused

CandelR (commented):
I think that in the folder workflow, if the backend returns an error because the frontend check fails, this won’t explain why the files have been skipped. Maybe we need to be more explicit and state why they haven’t been uploaded.

src/components/modals/AddModal/hooks/useFolderUpload.ts
```diff
@@ -131,8 +157,10 @@ export const useFolderUpload = ({ uploadAndCreateFileEntry }: { uploadAndCreateF
   const folderUploads = useAppSelector((state) => state.drive.folderUploads);
   const { limit } = useAppSelector((state) => state.storage);
   const usage = useAppSelector(storageSelectors.usage);
+  const maxUploadFileSize = 0;
```
CandelR: `maxUploadFileSize` is always 0. `filterOversizedFromTree` never will run correctly, always will return empty list. Correct this and add tests please :)

CandelR (changes requested):

CandelR (approved):

PR #458, opened 2026-05-19
[PB-6326]: bump @internxt/sdk to 1.16.4 and axios to 1.16.1

terrerox:
- Bump @internxt/sdk 1.13.2 → 1.16.4 to consume the typed `maxUploadFileSize` field on `getFileVersionLimits()`, replacing the local `as { maxUploadFileSize?: number }` casts in StorageService and useShareUpload.
  - Bump axios ^1.15.2 → ^1.16.1 to match the SDK's transitive axios, deduping the install and removing the `InternalAxiosRequestConfig` type mismatch in the rate-limit interceptor wiring.
  - Adjust call sites for SDK breaking changes:
    - `uploadMultipartFile` gained an `AbortSignal` parameter before `parts`; thread the existing abort signal in NetworkFacade, pass `undefined` in shareUploadService.
    - `Payments.createSession` was removed; delete the unused wrapper and import in PaymentService.


src/shareExtension/services/shareUploadService.ts
```diff
@@ -298,6 +298,7 @@ const shareUploadMultipartFile = async (context: UploadFileContext): Promise<{ f
         );
         return { hash: combinedPartsHashHex, parts: completedParts };
       },
+      undefined,
```
TamaraFinogina, line 301: why we need undefined here? Shouldn't it be an abort signal?

terrerox: This is the “Share to Internxt” entry point. When a user is in another app (Photos, Files, Safari, etc.) and taps the system Share button, Internxt appears as a destination. For this scenario, there is no abort signal unless it is required for a future feature

TamaraFinogina (approved):

CandelR (approved):

PR #461, opened 2026-05-20
[PB-6401] feature/Delete photos action

CandelR:
### Delete action with tombstone
When a backed-up local photo is trashed, instead of deleting its `asset_sync` row, we now write a `status = 'deleted'` tombstone. `PhotoDeduplicator.getAssetsToSync` fetches these IDs upfront and skips them, so the photo is never re-queued for upload after being explicitly deleted from the cloud.

  ### Restore to cloud
New "Upload to cloud" action in the more sheet, visible when all selected items are `local && not-backed`. It clears the tombstone (marks the asset pending) and triggers upload.

## Summary
 - `asset_sync.ts`: added `'deleted'` to status CHECK constraint, `deleted_at` column, `markDeleted` upsert, `getDeletedIds` query; `getPendingAssets` now excludes `'deleted'` rows
 - `photosLocalDB.ts`: `AssetSyncStatus` includes `'deleted'`; `deletedAt` field on entry/row types
 - `PhotoDeduplicator.ts`: fetches deleted IDs in parallel with synced entries and skips tombstoned assets in the deduplication loop
 - `PhotoActionsService.ts`: `local-backed` branch calls `markAssetDeleted` instead of `deleteAssetSync`; new `restoreToCloud` function
 - `photos/index.ts`: `runUploadThunk` accepts optional `{ bypassEnabled?: boolean }` for force uploads
 - `usePhotoActions.ts`: added `handleRestore` and dispatches `runUploadThunk({ bypassEnabled: true })` after restore and calls `reloadCloud` after both trash and restore
 - `photoUtils.ts`: new `isLocalItemNotBacked` to display reupload action
 - `MoreActionsBottomSheet.tsx`: added `onRestore` prop and "Upload to cloud" action, visible only when all selected items are `local && not-backed`
 - `PhotoPreviewScreen/index.tsx`: added delete modal

### DIFF
--- a/assets/lang/strings.ts
+++ b/assets/lang/strings.ts
@@ -253,9 +253,11 @@ const translations = {
             uploadingToCloud: 'Uploading to cloud…',
           },
           deleteModal: {
-            title: 'Delete item(s)',
-            message:
-              'These photos will be deleted from Internxt Drive on all your devices. They will remain in Trash for a limited time.',
+            title: (count: number) => (count === 1 ? 'Delete item' : 'Delete items'),
+            message: (count: number) =>
+              count === 1
+                ? 'This photo will be deleted from Internxt Drive on all your devices. It will remain in Trash for a limited time.'
+                : 'These photos will be deleted from Internxt Drive on all your devices. They will remain in Trash for a limited time.',
             confirm: 'Delete',
           },
         },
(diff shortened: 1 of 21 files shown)


TamaraFinogina (approved):
Sonar says you can remove as never from some tests

PR #464, opened 2026-05-27
[PB-6281]: add encrypted file upload flow with multipart support

terrerox:
- add bridge API methods/models for upload start, multipart finish, and file entry creation
- introduce encrypted uploader with SHA-256 part hashing and RIPEMD160 shard hash
- add upload token helpers in DocumentId and JSON parsing utilities
- add upload HTTP client config and Bouncy Castle RIPEMD160 dependency
- add crypto callback await helper and pending upload model

### DIFF
--- a/android/app/src/main/java/com/internxt/cloud/documents/api/model/CreateFileEntry.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/api/model/CreateFileEntry.kt
@@ -1,0 +1,15 @@
+package com.internxt.cloud.documents.api.model
+
+const val ENCRYPT_VERSION_AES03 = "Aes03"
+
+data class CreateFileEntry(
+    val fileId: String,
+    val type: String,
+    val size: Long,
+    val plainName: String,
+    val bucket: String,
+    val folderUuid: String,
+    val encryptVersion: String = ENCRYPT_VERSION_AES03,
+    val modificationTime: String? = null,
+    val creationTime: String? = null,
+)
--- a/android/app/src/main/java/com/internxt/cloud/documents/api/model/CreateFileEntry.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/api/model/CreateFileEntry.kt
@@ -1,6 +1,6 @@
 package com.internxt.cloud.documents.api.model
 
-const val ENCRYPT_VERSION_AES03 = "Aes03"
+const val ENCRYPT_VERSION_AES03 = "03-aes"
 
 data class CreateFileEntry(
     val fileId: String,
--- a/android/app/src/main/java/com/internxt/cloud/documents/crypto/CryptoServiceAwait.kt
+++ b/android/app/src/main/java/com/internxt/cloud/documents/crypto/CryptoServiceAwait.kt
@@ -1,0 +1,26 @@
+package com.internxt.cloud.documents.crypto
+
+import com.rncrypto.util.OnlyErrorCallback
+import java.io.IOException
+import java.util.concurrent.CompletableFuture
+import java.util.concurrent.ExecutionException
+
+/**
+ * Bridges `CryptoService.encryptFile` / `decryptFile` (callback-based) to a blocking call.
+ * Pass `runInBackground = false` to those methods — the caller is already on a background
+ * executor thread, so there's no point bouncing onto the rn-crypto thread pool just to
+ * block on it. Any [Throwable] surfaced by the callback is rethrown as [IOException]
+ * preserving the original cause.
+ */
+internal inline fun awaitCryptoService(
+    failureMessage: String,
+    invoke: (OnlyErrorCallback) -> Unit,
+) {
+    val done = CompletableFuture<Unit>()
+    invoke { err -> if (err != null) done.completeExceptionally(err) else done.complete(Unit) }
+    try {
+        done.get()
+    } catch (e: ExecutionException) {
+        throw (e.cause as? IOException) ?: IOException(failureMessage, e.cause)
+    }
+}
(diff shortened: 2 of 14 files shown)


android/app/src/main/java/com/internxt/cloud/documents/api/model/CreateFileEntry.kt
```diff
@@ -0,0 +1,15 @@
+package com.internxt.cloud.documents.api.model
+
+const val ENCRYPT_VERSION_AES03 = "Aes03"
```
TamaraFinogina: Shouldn't it be the same string as in the SDK? https://github.com/internxt/sdk/blob/master/src/drive/storage/types.ts#L204

terrerox: Right, i missed that. Thanks!

android/app/src/main/java/com/internxt/cloud/documents/crypto/CryptoServiceAwait.kt
```diff
@@ -0,0 +1,26 @@
...
+ * Pass `runInBackground = false` to those methods — the caller is already on a background
+ * executor thread, so there's no point bouncing onto the rn-crypto thread pool just to
+ * block on it. Any [Throwable] surfaced by the callback is rethrown as [IOException]
+ * preserving the original cause.
+ */
+internal inline fun awaitCryptoService(
+    failureMessage: String,
+    invoke: (OnlyErrorCallback) -> Unit,
+) {
+    val done = CompletableFuture<Unit>()
```
CandelR, line 19: Try to use coroutines, Kotlin and the entire Android ecosystem have moved to coroutines as the standard async model.`CompletableFuture.get()` blocks the thread while waiting, coroutines suspend instead, freeing the thread for other work, and cancellation is automatic when the coroutine scope is cancelled . Replace with `suspendCancellableCoroutine`

android/app/src/main/java/com/internxt/cloud/documents/upload/EncryptedFileUploader.kt
```diff
@@ -0,0 +1,257 @@
...
+            val length = (total - offset).coerceAtMost(partSize)
+            require(length > 0) { "Computed empty part for index $index" }
+            val body = fileRangeBody(tempEnc, offset, length, onProgress, baseSent = offset)
+            val request = Request.Builder().url(url).put(body).build()
+            var etag: String? = null
+            runCall(client, request, signal, active) { response ->
+                etag = response.header("ETag") ?: response.header("Etag")
+                    ?: throw IOException("Part ${index + 1} missing ETag in response")
+            }
+            parts.add(UploadedPart(partNumber = index + 1, etag = etag!!))
```
CandelR: do not force the no nullable values (!!), this could lead in unexpected errors. Return the resutl to etag directly

CandelR (changes requested):
@terrerox After reviewing the upload implementation and taking another look at the download implementation, I have noticed that the entire module (`EncryptedFileDownloader`,`EncryptedFileUploader`, `InternxtDocumentsProvider`) relies on blocking threads:
  - `call.execute()` blocks a thread for the duration of the transfer
  - `CompletableFuture.get()` blocks another thread during encryption
  - `openExecutor` exists solely to offload work from the binder thread, so it seems that waster a third thread. Opening a file ties up 3 threads, 2 of which do nothing but wait.

Migrating to coroutines solves this at the root:
  - `CryptoService` is pure native Java, so it can be wrapped directly with `suspendCancellableCoroutine`
  - `call.execute()` becomes `call.await()` via `enqueue` + `suspendCancellableCoroutine`
  - All the `AtomicReference` + `setOnCancelListener` + `OperationCanceledException` wiring disappears,  coroutine cancellation handles it automatically
  - `openExecutor` can be removed entirely, `runBlocking` at the boundary of `openDocument` is the appropriate bridge

If you’d like to give this a go, it might be worth creating a ticket to address this issue before continue with iOS tasks, as coroutines are the standard concurrency model in modern Kotlin for Android and would significantly simplify the cancellation logic

android/app/src/main/java/com/internxt/cloud/documents/crypto/CryptoServiceAwait.kt
terrerox: I will create another task to address this issue

terrerox:
> @terrerox After reviewing the upload implementation and taking another look at the download implementation, I have noticed that the entire module (`EncryptedFileDownloader`,`EncryptedFileUploader`, `InternxtDocumentsProvider`) relies on blocking threads:
> 
> * `call.execute()` blocks a thread for the duration of the transfer
> * `CompletableFuture.get()` blocks another thread during encryption
> * `openExecutor` exists solely to offload work from the binder thread, so it seems that waster a third thread. Opening a file ties up 3 threads, 2 of which do nothing but wait.
> 
> Migrating to coroutines solves this at the root:
> 
> * `CryptoService` is pure native Java, so it can be wrapped directly with `suspendCancellableCoroutine`
> * `call.execute()` becomes `call.await()` via `enqueue` + `suspendCancellableCoroutine`
> * All the `AtomicReference` + `setOnCancelListener` + `OperationCanceledException` wiring disappears,  coroutine cancellation handles it automatically
> * `openExecutor` can be removed entirely, `runBlocking` at the boundary of `openDocument` is the appropriate bridge
> 
> If you’d like to give this a go, it might be worth creating a ticket to address this issue before continue with iOS tasks, as coroutines are the standard concurrency model in modern Kotlin for Android and would significantly simplify the cancellation logic

Implemented coroutines [here](https://github.com/internxt/drive-mobile/pull/482)

CandelR (approved):

PR #465, opened 2026-05-27
[PB-6378] feature/Detect and mark photos deleted from cloud backup

CandelR:
Detect photos deleted from the cloud backup and mark them locally so the UI accurately reflects which photos are no longer backed up, and prevent them from being reuploaded automatically.

  ## Summary

  - **`reconcileCloudDeletions`** (`PhotoCloudBrowser.ts`): after fetching each month from Drive, compares found `remote_file_id`s against `asset_sync` and `cloud_asset` tables for that device/month. Any file absent in Drive is marked `cloud_deleted` in `asset_sync` and its `cloud_asset` row is removed.

  - **`reconcileDeletedMonths`** (`PhotoCloudBrowser.ts`): new pass that runs after all months are fetched. Queries the DB for months the device has uploaded (`cloud_asset` + `asset_sync synced`) and calls`reconcileCloudDeletions` with `foundIds` for any month that no longer exists in cloud Photos (this covers the case where an entire month folder is deleted).

  - **`cloud_deleted` status** added to `asset_sync` CHECK constraint. `PhotoDeduplicator.ts` skips assets with this status so they are never requeued unless the user edits them (in which case they are re-uploaded as new).

  - **`useLocalAssets`**: exposes `cloudDeletedIds` (separate from `syncedIds`) and reacts to `isFetchingCloudHistory` so the `CloudSlashIcon` appears automatically when the sync cycle finishes, without requiring the user to navigate away.

  - **`useCloudAssets`**: adds `sessionUploadedAssets` to the debounced reload effect so cloud-only items are 
  deduped correctly of an upload completing.

### DIFF
--- a/src/screens/PhotosScreen/hooks/useLocalAssets.ts
+++ b/src/screens/PhotosScreen/hooks/useLocalAssets.ts
@@ -12,6 +12,7 @@ export interface LocalAssetsResult {
   assets: MediaLibrary.Asset[];
   isLoading: boolean;
   syncedIds: Set<string>;
+  cloudDeletedIds: Set<string>;
   uploadingIdSet: Set<string>;
   loadNextPage: () => void;
   reload: () => Promise<void>;
@@ -21,13 +22,17 @@ export const useLocalAssets = (): LocalAssetsResult => {
   const [assets, setAssets] = useState<MediaLibrary.Asset[]>([]);
   const [isLoading, setIsLoading] = useState(true);
   const [syncedIds, setSyncedIds] = useState<Set<string>>(new Set());
+  // TODO: Reserved for show a distinct icon for cloud-deleted vs never-backed assets. Remove if finally will be the same.
+  const [cloudDeletedIds, setCloudDeletedIds] = useState<Set<string>>(new Set());
 
   const cursorRef = useRef<string | undefined>(undefined);
   const hasMoreRef = useRef(true);
   const isLoadingMoreRef = useRef(false);
   const appStateRef = useRef(AppState.currentState);
 
-  const { syncStatus, uploadingAssetIds, sessionUploadedAssets } = useAppSelector((state) => state.photos);
+  const { syncStatus, uploadingAssetIds, sessionUploadedAssets, isFetchingCloudHistory } = useAppSelector(
+    (state) => state.photos,
+  );
 
   const uploadingIdSet = useMemo(() => new Set(uploadingAssetIds), [uploadingAssetIds]);
 
@@ -88,7 +93,17 @@ export const useLocalAssets = (): LocalAssetsResult => {
     const assetIds = assets.map((a) => a.id);
     await photosLocalDB.init();
     const entries = await photosLocalDB.getSyncedEntries(assetIds);
-    setSyncedIds(new Set(entries.keys()));
+    const synced = new Set<string>();
+    const cloudDeleted = new Set<string>();
+    for (const [id, info] of entries) {
+      if (info.status === 'cloud_deleted') {
+        cloudDeleted.add(id);
+      } else {
+        synced.add(id);
+      }
+    }
+    setSyncedIds(synced);
+    setCloudDeletedIds(cloudDeleted);
   }, [assets]);
 
   const reloadFromStart = useCallback(async () => {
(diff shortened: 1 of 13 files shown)


src/screens/PhotosScreen/hooks/useLocalAssets.ts
```diff
@@ -21,13 +22,17 @@ export const useLocalAssets = (): LocalAssetsResult => {
   const [assets, setAssets] = useState<MediaLibrary.Asset[]>([]);
   const [isLoading, setIsLoading] = useState(true);
   const [syncedIds, setSyncedIds] = useState<Set<string>>(new Set());
+  // TODO: Reserved for show a distinct icon for cloud-deleted vs never-backed assets. Remove if finally will be the same.
+  const [cloudDeletedIds, setCloudDeletedIds] = useState<Set<string>>(new Set());
```
TamaraFinogina, line 26: Btw, why would we need a distinction? They are not backed up all the same, no?

src/services/photos/database/photosLocalDB.ts
```diff
@@ -256,6 +257,47 @@ class PhotosLocalDB {
...
+
+  async getSyncedMonths(): Promise<{ year: number; month: number }[]> {
+    return sqliteService.getAllAsync<{ year: number; month: number }>(
+      DB_NAME,
+      assetSyncTable.statements.getSyncedMonths,
+    );
+  }
+
+  async getCloudAssetRemoteIdsByDeviceAndMonth(deviceId: string, year: number, month: number): Promise<Set<string>> {
+    const startMs = new Date(year, month - 1, 1).getTime();
```
TamaraFinogina, line 291: why month -1? Are you dividing them based on the 1st of the last month till 1st of this month?

src/services/photos/PhotoCloudBrowser.spec.ts
```diff
@@ -372,18 +384,253 @@ describe('PhotoCloudBrowser.syncAllHistory', () => {
...
+    const month = makeFolder('m-uuid', '06');
+    const day = makeFolder('day-uuid', '15');
+    mockPhotosLocalDB.getCloudFetchCacheAge.mockResolvedValue(null);
+    mockPhotosLocalDB.getCloudAssetRemoteIdsByDeviceAndMonth.mockResolvedValue(new Set());
+    mockFolderService.getFolderFolders
+      .mockResolvedValueOnce({ folders: [device] })
+      .mockResolvedValueOnce({ folders: [year] })
+      .mockResolvedValueOnce({ folders: [month] })
+      .mockResolvedValueOnce({ folders: [day] });
+    (mockFolderService.getFolderContentByUuid as jest.Mock).mockResolvedValueOnce({ files: [] });
```
TamaraFinogina, line 500: Sonar says as jest.Mock is not needed

src/services/photos/PhotoCloudBrowser.ts
```diff
@@ -173,6 +185,67 @@ class PhotoCloudBrowserService {
...
+
+    if (currentDeviceId && deviceId === currentDeviceId) {
+      const knownFromSync = await this.localDB.getSyncedRemoteIdsByCreationMonth(year, month);
+      for (const id of knownFromSync) knownIds.add(id);
+    }
+
+    let removedCount = 0;
+    for (const knownId of knownIds) {
+      if (!foundIds.has(knownId)) {
+        await this.localDB.markCloudDeleted(knownId);
```
TamaraFinogina, line 207: Maybe it can be one DB request?

src/screens/PhotosScreen/hooks/useLocalAssets.ts
CandelR: Yes, but one won’t be automatically re-uploaded because the user has deleted their copy, whereas the other will. That’s why I left the comment pending confirmation from the design team as to whether we want to distinguish between them or not

src/services/photos/database/photosLocalDB.ts
CandelR: months in Date goes from 0 to 11 😩

TamaraFinogina (approved):

src/services/photos/PhotoCloudBrowser.ts
CandelR: It cannot be a single SQL statement, as it involves two different tables, but we can expose it as a single function wrapping both in a transaction

src/services/photos/PhotoCloudBrowser.spec.ts
CandelR: I can't see the message on SoundCloud. Anyway, if I remove it, it gives an error because it doesn't have the return type corrctly, but is not necessary all the fields for the test

PR #467, opened 2026-05-27
[PB-6081] feature/Added actions to more button in preview screen

CandelR:
Connect the preview screen's action bar to the existing `MoreActionsBottomSheet`, add an Info entry that opens `MetadataPanel`, and visually disable Export and Delete when the current item is not backed up.

## Summary

  - **`MoreActionsBottomSheet`**: added optional `onInfo` prop. When provided, renders an "Info" `ActionRow` as
  the first item in the sheet

  - **`PreviewCarousel`**: added `isSynced: boolean` prop and `disabled?: boolean` to `ActionButton`

  - **`PhotoPreviewScreen`**: rewired `onMore` calls to open`MoreActionsBottomSheet` instead of `MetadataPanel` directly. `MetadataPanel` now opens via the Info row in the sheet.

  - duplicate action removed: dead code

### DIFF
--- a/src/screens/PhotoPreviewScreen/index.tsx
+++ b/src/screens/PhotoPreviewScreen/index.tsx
@@ -10,6 +10,8 @@ import { photoActionsService } from 'src/services/photos/PhotoActionsService';
 import { PhotoAssetFetchService } from 'src/services/photos/PhotoAssetFetchService';
 import { RootStackScreenProps } from 'src/types/navigation';
 import { useTailwind } from 'tailwind-rn';
+import MoreActionsBottomSheet from '../PhotosScreen/components/MoreActionsBottomSheet';
+import { isItemBacked } from '../PhotosScreen/utils/photoUtils';
 import { MetadataPanel } from './components/MetadataPanel';
 import { PreviewCarousel } from './components/PreviewCarousel';
 import { PreviewHeader } from './components/PreviewHeader';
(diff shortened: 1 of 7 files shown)


TamaraFinogina:
@CandelR Here, it also doesn't have coverage. Does mobile have coverage enabled? Because if so, it's broken

src/screens/PhotoPreviewScreen/index.tsx
```diff
@@ -117,8 +120,51 @@ export const PhotoPreviewScreen = ({ route }: Props): JSX.Element => {
...
+    try {
+      await photoActionsService.saveToDevice(item, controller.signal);
+    } catch (error) {
+      logger.error(`[PhotoPreview] Save failed: ${error}`);
+    }
+  }, [items, currentIndex]);
+
+  const handleRestore = useCallback(async () => {
+    const item = items[currentIndex];
+    if (!item || item.type !== 'local') {
```
TamaraFinogina, line 153: Sinar wants something like `if (item?.type !== 'local')` here

src/screens/PhotoPreviewScreen/index.tsx
```diff
@@ -150,11 +196,30 @@ export const PhotoPreviewScreen = ({ route }: Props): JSX.Element => {
...
+        onClose={() => setIsMoreActionsOpen(false)}
+        onInfo={() => {
+          setIsMoreActionsOpen(false);
+          setMetadataOpen(true);
+        }}
+        onExport={handleExport}
+        onCopy={handleCopy}
+        onSave={handleSave}
+        onFavorite={() => {
+          logger.info('[PhotoPreview] Favorite: not implemented');
```
TamaraFinogina, line 217: Why is the action here then?

CandelR: It was planned, but has been postponed

src/screens/PhotoPreviewScreen/index.tsx
CandelR: I saw it, but I prefer it this way; I think it looks clearer. What do you think?

TamaraFinogina (approved):

larryrider (approved):

PR #473, opened 2026-05-29
[v1.9.1] - Release 1.9.1

CandelR:


src/store/slices/storage/index.ts
```diff
@@ -8,13 +8,19 @@ import { driveThunks } from '../drive';
...
 }
 
 const initialState: StorageState = {
   limit: 0,
   totalUsage: 0,
+  maxUploadFileSize: 0,
+  maxUploadFileSizeFetchedAt: 0,
 };
 
+const MAX_UPLOAD_FILE_SIZE_TTL_MS = 60_000;
```
TamaraFinogina, line 22: Maybe all constants should go to a separate file. After a while, it will be hard to locate them

src/types/drive/operations.ts
```diff
@@ -75,6 +75,4 @@ export interface DownloadingFile {
 /**
  * CONSTANTS
  */
-const GB = 1024 * 1024 * 1024;
-export const UPLOAD_FILE_SIZE_LIMIT = 5 * GB;
 export const DRIVE_DB_NAME = 'drive.db';
```
TamaraFinogina, line 78: this one too

TamaraFinogina (dismissed):

src/store/slices/storage/index.ts
CandelR: If it’s only used in that file, I think is not necessary to move it

neuffer:
PLEASE add mounting the drive in a usable place in the Android FS.
Currently you could just as well only use the web interface and not 
have an Android client at all. Normal applications can not access the drive. 
Can it be so hard to add a directory on this level?
Others like pCloud can do it as well, why don't you? 
signal-2026-06-02-100334

CandelR:
> PLEASE add mounting the drive in a usable place in the Android FS. Currently you could just as well only use the web interface and not have an Android client at all. Normal applications can not access the drive. Can it be so hard to add a directory on this level? Others like pCloud can do it as well, why don't you? signal-2026-06-02-100334

Hey neuffer, please don’t make a habit of leaving Product questions or requests in the PRs 🙏 Try to use the support channel.
The integration of Internxt Drive with Android’s SAF is already underway, so the feature will be released sooner or later.

TamaraFinogina (approved):

PR #480, opened 2026-06-05
[PB-6282]: refresh file picker after uploads and folder creation

terrerox:
Add a native signaling bridge from React Native to the Android SAF provider.
Reuse provider cache invalidation for RN-originated parent folder changes.
Signal after successful file uploads and folder creation, with iOS/no-module no-op behavior.
Add Jest coverage for the wrapper, upload signaling, and folder signaling.

### DIFF
--- a/src/services/drive/file/utils/uploadFileUtils.signaling.spec.ts
+++ b/src/services/drive/file/utils/uploadFileUtils.signaling.spec.ts
@@ -65,28 +65,22 @@ describe('uploadSingleFile — picker signaling', () => {
   });
 
   it('when a file upload succeeds, then it signals the destination folder uuid', async () => {
-    // given
     const file = buildUploadingFile({ parentUuid: 'destination-folder-uuid' });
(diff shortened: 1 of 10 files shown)


src/services/drive/file/utils/uploadFileUtils.signaling.spec.ts
```diff
@@ -0,0 +1,86 @@
...
+});
+
+describe('uploadSingleFile — picker signaling', () => {
+  const dispatch = jest.fn() as unknown as Dispatch<Action>;
+
+  beforeEach(() => {
+    mockNotifyParentChanged.mockClear();
+  });
+
+  it('when a file upload succeeds, then it signals the destination folder uuid', async () => {
```
xabg2: Would be better to use `test` when describing the tests with When..., then...

xabg2 (approved):

CandelR (approved):

PR #489, opened 2026-06-10
[PB-5919]: register File Provider domain on login/logout

terrerox:
Sync auth credentials to a shared App Group Keychain so the File Provider extension can read them; register the com.internxt.drive domain on login and call removeAllDomains on logout so the provider appears/disappears in Files.app. Bump InternxtFileProvider MARKETING_VERSION 1.0 -> 1.9.0 (Debug+Release) so the extension version matches the host or iOS won't load it. Device-verified on a real iPhone.

### DIFF
--- a/ios/Internxt/AppDelegate.swift
+++ b/ios/Internxt/AppDelegate.swift
@@ -1,7 +1,6 @@
 import Expo
 import React
 import ReactAppDependencyProvider
-import Security
 
 @UIApplicationMain
 public class AppDelegate: ExpoAppDelegate {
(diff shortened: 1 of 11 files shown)


ios/Internxt/FileProviderDomainManager.swift
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  static func registerDomain(completion: @escaping (Error?) -> Void) {
+    NSFileProviderManager.add(domain) { error in
+      completion(isAlreadyExists(error) ? nil : error)
+    }
+  }
+
+  static func unregisterDomain(completion: @escaping (Error?) -> Void) {
+    NSFileProviderManager.removeAllDomains { removeError in
```
patricioxavier8, line 20: It's better to use NSFileProviderManager.remove, and review the available file types to see which one best suits your implementation.

patricioxavier8 (changes requested):

ios/Internxt/FileProviderDomainManager.swift
terrerox: I reviewed the available removal modes. On iOS the only one is` .removeAll`. The `.preserveDirtyUserData` and `.preserveDownloadedUserData` are macOS-only (the build fails with `'preserveDirtyUserData' is unavailable in iOS`). Since `.removeAll` is the only mode iOS offers and we register a single domain, `remove(_:mode: .removeAll)` would be functionally  identical to removeAllDomains at runtime. So I kept removeAllDomains for simplicity, same behavior, less code. WDYT?

patricioxavier8: You're right, I hadn't noticed that the different removal modes are only available on macOS. Given that, I don't see any issue with keeping removeAllDomains() for now. The main consideration would be the overall design and whether we plan to support multiple domains in the future. If we ever move to multiple registered domains, using remove(_:mode:) explicitly might provide more flexibility and avoid unintentionally removing all domains.

patricioxavier8 (approved):

ios/Internxt/AppDelegate.swift
```diff
@@ -44,75 +43,7 @@ public class AppDelegate: ExpoAppDelegate {
...
-  private func deleteFromSharedKeychain(key: String, accessGroup: String) {
-    let query: [String: Any] = [
-      kSecClass as String: kSecClassGenericPassword,
-      kSecAttrService as String: "app:no-auth",
-      kSecAttrGeneric as String: Data(key.utf8),
-      kSecAttrAccount as String: Data(key.utf8),
-      kSecAttrAccessGroup as String: accessGroup,
-    ]
-    SecItemDelete(query as CFDictionary)
+    SharedAuthKeychain.syncFromPrivateKeychain()
```
CandelR, line 46: good clean up @terrerox

src/services/native/InternxtAuthCredentialsModule.ts
```diff
@@ -18,7 +18,7 @@ interface NativeBridge {
 }
 
 const bridge: NativeBridge | undefined =
-  Platform.OS === 'android' ? NativeModules.InternxtAuthCredentialsModule : undefined;
+  Platform.OS === 'android' || Platform.OS === 'ios' ? NativeModules.InternxtAuthCredentialsModule : undefined;
```
CandelR, line 21: nit: this check is not necessary anymore :), could remove in future changes

CandelR (approved):

PR #492, opened 2026-06-12
[PB-5921]: File Provider folder browsing (read-only enumeration)

terrerox:
Implements read-only browsing of Drive in the iOS Files.app via the `InternxtFileProvider` extension.
  
### What this delivers
  - **Root + subfolder enumeration** — lists folders and files from the Drive API
    and navigates into subfolders.
  - **Correct names** — `plainName` + extension.
  - **Pagination** — 50 items/page via `NSFileProviderPage` as the cursor
    (folders first, then files, `plainName ASC`), so large folders show all items.
  - **Item metadata** — size, creation/modification dates, and per-kind
    materialization flags so items render as normal cloud items (folders
    materialized; files cloud-downloadable) instead of greyed/locked placeholders.
  - **Auth/offline states** — missing credentials → `notAuthenticated`
    (drives the system "Sign in to Internxt Drive" prompt); network errors
    surface as errors without crashing.
  - **Drive API wiring** — `DriveAPIFactory` builds an authenticated client from
    the shared App Group keychain; adds a shared `driveBaseUrl` key.
  - **Type-prefixed item identifiers** (`f:`/`d:`) so a container identifier
    round-tripped by Files.app resolves back to a Drive folder uuid.

### DIFF
--- a/ios/InternxtFileProvider/FileProviderItem.swift
+++ b/ios/InternxtFileProvider/FileProviderItem.swift
@@ -6,40 +6,188 @@
 //
 
 import FileProvider
+import InternxtSwiftCore
 import UniformTypeIdentifiers
 
 class FileProviderItem: NSObject, NSFileProviderItem {
 
-    // TODO: implement an initializer to create an item from your extension's backing model
-    // TODO: implement the accessors to return the values from your extension's backing model
-    
-    private let identifier: NSFileProviderItemIdentifier
-    
-    init(identifier: NSFileProviderItemIdentifier) {
-        self.identifier = identifier
+  private let identifier: NSFileProviderItemIdentifier
+  private let parent: NSFileProviderItemIdentifier
+  private let name: String
+  private let kind: DriveItemKind
+  private let fileExtension: String?
+  private let updatedAt: String?
+  private let createdAt: String?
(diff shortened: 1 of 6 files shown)


patricioxavier8 (approved):

ios/InternxtFileProvider/FileProviderItem.swift
```diff
@@ -6,40 +6,188 @@
...
+  }()
+
+  private static func parseISO8601(_ value: String?) -> Date? {
+    guard let value = value, !value.isEmpty else { return nil }
+    if let date = iso8601Formatter.date(from: value) {
+      return date
     }
+    let fallback = ISO8601DateFormatter()
+    fallback.formatOptions = [.withInternetDateTime]
+    return fallback.date(from: value)
```
CandelR: I think that `fallback` creates a new `ISO8601DateFormatter` on every call, and this is executed for each item in the enumeration. Cache it as a static let, just like `iso8601Formatter` above, do not mutate an instance shared between the two formats because it is called from concurrent Task {} in the enumerator.
Also, I would rename `parseISO8601` to `parseDate`

CandelR (approved):

PR #500, opened 2026-06-16
[PB-5923]: create folders and upload encrypted files via File Provider

terrerox:
Implement the write path in the NSFileProviderReplicatedExtension so the Files.app can create folders and upload files to Drive. createItem routes folder creation to the Drive API and file uploads through InternxtSwiftCore encryption plus network upload, then registers the file and returns the created NSFileProviderItem. Adds rename support via modifyItem and friendly non-crashing errors for offline/unauthenticated/generic failures.

### DIFF
--- a/ios/InternxtFileProvider/FileProviderDownloadService.swift
+++ b/ios/InternxtFileProvider/FileProviderDownloadService.swift
@@ -1,0 +1,37 @@
+//
+//  FileProviderDownloadService.swift
+//  InternxtFileProvider
+//
+//  Created by Ramon Candel on 9/4/26.
+//
+
+import FileProvider
+import InternxtSwiftCore
+
+struct FileProviderDownloadService {
+    let driveAPI: DriveAPI
+    let networkFacade: NetworkFacade
+
+    func fetchContents(
+        uuid: String,
+        progressHandler: @escaping (Double) -> Void
+    ) async throws -> (url: URL, meta: GetFileMetaByIdResponse) {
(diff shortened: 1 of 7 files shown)


ios/InternxtFileProvider/FileProviderExtension.swift
patricioxavier8, line 1: Consider breaking this file apart as it grows — right now `FileProviderExtension` is handling protocol conformance, upload logic, bucket resolution, and error mapping all in one place. As we add move/trash/versioning this will get harder to test and maintain

patricioxavier8 (changes requested):

patricioxavier8 (approved):

ios/InternxtFileProvider/FileProviderDownloadService.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  FileProviderDownloadService.swift
+//  InternxtFileProvider
+//
+//  Created by Ramon Candel on 9/4/26.
```
CandelR, line 5: Hey @terrerox, this is autogenerated by Xcode when create a file, no need to added it to new files 😵‍💫

CandelR (approved):

ios/InternxtFileProvider/FileProviderDownloadService.swift
terrerox: Got it, i will modify it in another PR

PR #509, opened 2026-06-19
[_] bugfix/Improve asset changes detection

CandelR:
Fix locally deleted photos not appearing as cloud-only, add real-time gallery sync, and improve photos startup performance

  - **`useLocalAssets`**: replaced AppState-only reconcile with `MediaLibrary.addListener` for real time change detection while the app is open. iOS incremental events (inserted/deleted/updated assets) are applied without a fetch. Non-incremental events (Android, bulk changes) fall back to a debounced `reconcileOnForeground`. AppState listener kept as safety net for frozen process resume.
  - **`photosLocalDB` / `asset_sync`**: added `cleanupOrphanedAssetSync(localAssetIds)` that removes `asset_sync` entries for assets no longer on device and `getAllTrackedAssetIds` to support it.
  - **`runDiscoveryThunk`**: calls `cleanupOrphanedAssetSync` after each scan so historical local deletions are reconciled every backup cycle.
  - **`usePhotosTimeline`**: watches `localDeletionDetectedCount` and calls `reloadCloud` whenever local deletions are detected, so cloud copies surface as cloud-only immediately.
  - **`PhotoAssetScanner`**: increased scan page size from 200 → 1000 to reduce the number of async iterations on large libraries.
  - **`payments/index`**: added 24 h TTL cache (AsyncStorage) for the photos access check to avoid a network call on every cold start.
  - **`App`**: added `useKeepAwake()` to prevent the screen from sleeping during backup, and dispatches `loadFileLimitsThunk` on init.
  - **`PreviewHeader`**: upload arrow icon now animates while a photo is being uploaded.
  - **`PhotoItem`**: fixed skeleton animation to use `useNativeDriver: true`.
  - **`useDebouncedValue`**: new generic debounce hook used by `useLocalAssets` to coalesce burst queries.

### DIFF
--- a/src/hooks/useDebouncedValue.spec.ts
+++ b/src/hooks/useDebouncedValue.spec.ts
@@ -1,0 +1,85 @@
+import { act, renderHook } from '@testing-library/react-native';
+import useDebouncedValue from './useDebouncedValue';
+
+jest.useFakeTimers();
+
+describe('useDebouncedValue', () => {
+  afterEach(() => {
+    jest.clearAllTimers();
+  });
+
+  test('when the hook mounts, then the initial value is returned immediately', () => {
+    const { result } = renderHook(() => useDebouncedValue('hello', 300));
+
+    expect(result.current).toBe('hello');
+  });
+
+  test('when the value changes, then the debounced value does not update before the delay elapses', () => {
+    const { result, rerender } = renderHook(({ value }) => useDebouncedValue(value, 300), {
+      initialProps: { value: 'first' },
+    });
+
+    rerender({ value: 'second' });
+    act(() => {
+      jest.advanceTimersByTime(299);
+    });
+
+    expect(result.current).toBe('first');
+  });
+
+  test('when the delay elapses after a value change, then the debounced value updates', () => {
+    const { result, rerender } = renderHook(({ value }) => useDebouncedValue(value, 300), {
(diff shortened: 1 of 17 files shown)


src/hooks/useDebouncedValue.spec.ts
```diff
@@ -0,0 +1,85 @@
+import { act, renderHook } from '@testing-library/react-native';
+import useDebouncedValue from './useDebouncedValue';
+
+jest.useFakeTimers();
```
larryrider: maybe this statement is better to be included in a beforeAll function instead of here 🤔

CandelR: In principle, it has the same effect, but if you’re saying it for the sake of consistency, that’s almost better. I’ll change it

src/hooks/useDebouncedValue.spec.ts
```diff
@@ -0,0 +1,84 @@
+import { act, renderHook } from '@testing-library/react-native';
+import useDebouncedValue from './useDebouncedValue';
+
+describe('useDebouncedValue', () => {
+  afterEach(() => {
+    jest.clearAllTimers();
+    jest.useFakeTimers();
```
larryrider: better before

larryrider (approved):

PR #511, opened 2026-06-22
[_] bugfix/Design fixes

CandelR:
Added design adjustments to the selection UI to fit design and a fix for the Security modal on the Account screen.
Also added (on iOS only) in select mode, drag across the photo grid to select a range of photos in one gesture.

 ## Summary

  - **Photos selection UI**: removed the standalone`SelectionHeader` component — cancel button and selection count are now part of `PhotosHeader`, which handles both normal and select-mode states. Unselected items no longer show an empty checkbox. Removed date group section headers from the timeline grid.

  - **Swipe-to-select gesture (iOS)**: in select mode, hold and drag across the grid to select or deselect a range of photos in a single gesture. The anchor is captured at touch-down (before pan activation) so the first cell under the finger is always included. Dragging from an already selected photo deselects the range. Auto-scrolls when the finger reaches the top or bottom edge. Disabled on Android due to an unresolvable conflict between `Gesture.Pan` and `FlashList` scroll.

  - **SecurityModal**: fixed incorrect rendering on the Account screen


src/screens/PhotosScreen/components/GroupHeader/PhotosGroupHeader.tsx
```diff
@@ -102,41 +104,48 @@ const PhotosGroupHeader = memo(
...
+                    statusColor={statusColor}
+                    onResumePress={onResumePress}
+                  />
+                )}
+                {syncStatus.type === 'paused-storage-full' && <GroupHeaderPausedStorageFull dangerColor={dangerColor} />}
+                {syncStatus.type === 'paused-no-wifi' && <GroupHeaderPausedNoWifi color={statusColor} />}
+                {syncStatus.type === 'paused-no-connection' && <GroupHeaderPausedNoConnection color={statusColor} />}
+                {syncStatus.type === 'completed' && <GroupHeaderCompleted color={labelColor} />}
+                {syncStatus.type === 'upload-error' && (
+                  <GroupHeaderUploadError count={syncStatus.count} color={statusColor} onPress={onRetryPress} />
```
larryrider: Maybe its better to use a switch instead of a `{syncStatus.type === 'upload-error' && ....` for every component 🤔

larryrider (approved):

PR #518, opened 2026-07-01
[BR-2158]: decode percent-encoded URIs in document upload and thumbnail paths

terrerox:
Uploads of files whose names contain spaces or non-ASCII characters failed with ENOENT because percent-encoded cache URIs were passed unmodified to RNFS.copyFile and to the native thumbnail generator. Decode the source URI before copy (copyFileFromEncodedUri) and the native-returned thumbnail URI (fromFileUri); drop the stale manual %20 encoding. Device-verified on Android.

### DIFF
--- a/src/services/common/uri/uriHelpers.ts
+++ b/src/services/common/uri/uriHelpers.ts
@@ -3,7 +3,7 @@ export const FILE_URI_PREFIX = 'file://';
 export const toFileUri = (path: string): string => {
   if (path.startsWith(FILE_URI_PREFIX)) return path;
   const absolutePath = path.startsWith('/') ? path : `/${path}`;
-  return `${FILE_URI_PREFIX}${encodeURI(decodeURIComponent(absolutePath))}`;
+  return `${FILE_URI_PREFIX}${encodeURI(decodeUriSafely(absolutePath))}`;
 };
 
 export const decodeUriSafely = (uri: string): string => {
--- a/src/services/common/uri/uriHelpers.ts
+++ b/src/services/common/uri/uriHelpers.ts
@@ -6,5 +6,13 @@ export const toFileUri = (path: string): string => {
   return `${FILE_URI_PREFIX}${encodeURI(decodeURIComponent(absolutePath))}`;
 };
 
-export const stripFileUri = (path: string): string =>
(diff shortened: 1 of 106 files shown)


TamaraFinogina:
@terrerox This branch is full of conflicts now; they should be resolved first

src/services/common/uri/uriHelpers.ts
```diff
@@ -3,8 +3,16 @@ export const FILE_URI_PREFIX = 'file://';
...
-  path.startsWith(FILE_URI_PREFIX) ? decodeURIComponent(path.slice(FILE_URI_PREFIX.length)) : path;
+export const decodeUriSafely = (uri: string): string => {
+  try {
+    return decodeURIComponent(uri);
+  } catch {
+    return uri;
+  }
+};
+
+export const fromFileUri = (uri: string): string =>
```
CandelR, line 17: `fromFileUri`name not left clear what it does

src/services/drive/file/utils/uploadFileUtils.ts
```diff
@@ -173,6 +175,10 @@ export function isFileEmpty(file: { size: number }): boolean {
   return file.size === 0;
 }
 
+export async function copyFileFromEncodedUri(sourceUri: string, destPath: string): Promise<void> {
```
CandelR, line 178: I think this has little to do with upload, it would be better to move it to another utils file

src/services/common/uri/uriHelpers.ts
```diff
@@ -3,8 +3,16 @@ export const FILE_URI_PREFIX = 'file://';
 export const toFileUri = (path: string): string => {
   if (path.startsWith(FILE_URI_PREFIX)) return path;
   const absolutePath = path.startsWith('/') ? path : `/${path}`;
-  return `${FILE_URI_PREFIX}${encodeURI(decodeURIComponent(absolutePath))}`;
+  return `${FILE_URI_PREFIX}${encodeURI(decodeUriSafely(absolutePath))}`;
```
CandelR, line 6: `encodeURI` doesn't re-encode # or ? after `decodeUriSafely` decodes them, so Q1%23report.pdf becomes file:///cache/Q1#report.pdf and the thumbnail native modules truncate the path at #.
I think that would be good to add tests for these cases in the new spec to prevent any regression.

PR #525, opened 2026-07-08
[PB-6100] bugfix/Handle offline state

CandelR:
Fix crash on offline mode by making driveItems selector safe when bucket is missing

### DIFF
--- a/src/store/slices/drive/index.ts
+++ b/src/store/slices/drive/index.ts
@@ -605,6 +605,8 @@ export const driveSlice = createSlice({
   },
 });
 
+let hasLoggedMissingBucket = false;
+
 export const driveSelectors = {
   absolutePath: (state: RootState) => {
     return state.drive.navigationStack.reduce((result, item) => result + item.name + '/', '/');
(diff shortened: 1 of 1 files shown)


src/store/slices/drive/index.ts
```diff
@@ -618,7 +620,14 @@ export const driveSelectors = {
...
 
-    if (!bucket) throw new Error('Bucket not found');
+    if (!bucket) {
+      if (!hasLoggedMissingBucket) {
+        hasLoggedMissingBucket = true;
+        logger.warn('[Drive] driveItems selector evaluated with no bucket, returning empty result');
+      }
+      return { uploading: [], items: [] };
+    }
+    hasLoggedMissingBucket = false;
```
larryrider, line 630: if the idea is to only log 'no bucket' one time at all, then what is the use case of this line `hasLoggedMissingBucket = false;`? 🤔

CandelR: It is for rearm the log in case that the missing bucket was trasitory for an specific situation

larryrider (approved):

PR #526, opened 2026-07-09
[_] feature/Updated lib and token refresh period

CandelR:
- Updated lib in order to manage token expiration with same rules that other internxt projects

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "@dr.pogodin/react-native-fs": "2.36.2",
     "@expo/config-types": "^47.0.0",
     "@hookform/resolvers": "^2.9.1",
-    "@internxt/lib": "1.5.0",
+    "@internxt/lib": "1.5.1",
     "@internxt/mobile-sdk": "^0.3.1",
     "@internxt/rn-crypto": "0.1.15",
     "@internxt/sdk": "1.17.3",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -33,7 +33,7 @@
     "@dr.pogodin/react-native-fs": "2.36.2",
     "@expo/config-types": "^47.0.0",
     "@hookform/resolvers": "^2.9.1",
-    "@internxt/lib": "^1.4.1",
+    "@internxt/lib": "1.5.0",
```
larryrider: use the latest 1.5.1

larryrider (changes requested):

src/services/AuthService.ts
```diff
@@ -1,12 +1,12 @@
 import { logger } from '@internxt-mobile/services/common';
+import { auth, TokenStatus } from '@internxt/lib';
 import { internxtMobileSDKConfig } from '@internxt/mobile-sdk';
 import { TwoFactorAuthQR } from '@internxt/sdk';
 import { StorageTypes } from '@internxt/sdk/dist/drive';
 import { UserSettings } from '@internxt/sdk/dist/shared/types/userSettings';
 import * as bip39 from '@scure/bip39';
 import { wordlist } from '@scure/bip39/wordlists/english.js';
 import EventEmitter from 'events';
-import { jwtDecode } from 'jwt-decode';
```
larryrider, line 9: the `jwt-decode` dependency could be also removed from the package.json, right? 🤔

larryrider (changes requested):

larryrider (approved):

PR #529, opened 2026-07-11
[PB-6482]: reliably upload iCloud-offloaded camera-roll originals

terrerox:
- expo-image-picker (SDK 54) materialized all selected originals at pick time, all-or-nothing; an iCloud-offloaded original that couldn't download rejected the whole selection with PHPhotosErrorDomain 3164, which was swallowed silently (no upload, no error).
- Replace the iOS camera-roll selection with a native PHPickerViewController (PhotoPicker.pickAssets) that returns asset identifiers only, then download each original via the existing native PHAssetExport (isNetworkAccessAllowed) with retry and per-asset failure isolation; surface a user-facing error instead of swallowing.
- Wire PHAssetExportModule.swift/.m (which also host the new PhotoPicker module) into the Internxt app target in the Xcode project — they were previously not referenced, so the native modules never compiled in.
- Add cameraRollAssets service (pickPhotoAssets/exportPhAsset/withRetry) and i18n strings.


CandelR (changes requested):
I’d say there’s an official Expo fix for this: https://github.com/expo/expo/pull/40697. 
Is there anything else that justifies using the custom picker? After all, this means extra code to maintain and potential bugs, for something that’s already included in the Expo library

terrerox:
> I’d say there’s an official Expo fix for this: [expo/expo#40697](https://github.com/expo/expo/pull/40697). Is there anything else that justifies using the custom picker? After all, this means extra code to maintain and potential bugs, for something that’s already included in the Expo library

Good catch, but that fix only landed in SDK 55 and we're on 54 (no backport). Upgrading brings RN 0.83 and a bunch of things to re-check, so I think it deserves its own ticket.
Even on 55, the picker still fails the whole selection if one iCloud asset can't download, so we'd need the per-asset retry anyway. I can leave a TODO to revisit after the upgrade.

PR #533, opened 2026-07-13
Release 1.11.0- Native file-system integration

terrerox:
## Release 1.11.0 — Native file-system integration (Android SAF + iOS File Provider)

This release makes Internxt Drive available directly in the operating system's file pickers and file managers on both platforms, with end-to-end encryption preserved throughout.

### Android — Storage Access Framework (DocumentsProvider)

- **New `InternxtDocumentsProvider`** that exposes Drive in the system file picker and Files apps, shown/hidden based on session state.
- **Browse**: list folder contents with document row caching, MIME type resolution, and UUID-encoded document IDs.
- **Download & open**: files are downloaded and decrypted natively (new `InternxtApiClient` using the `/info` download-links endpoint, `FileKeyDeriver` backed by `CryptoService`, dedicated HTTP clients with coroutine-based IO).
- **Upload**: full encrypted upload pipeline from `createDocument`, including multipart uploads, a foreground service with progress notifications and cancel support, and picker refresh afteion.
- **File management**: create folders, rename, move, and trash directly from the SAF picker.
- **Auth bridge**: new `InternxtAuthCredentialsModule` syncs credentials (including mnemonic, root folder UUID, and client name/version) from
React Native to the native layer.
- Dependency bumps: OkHttp 5.3.2, security-crypto 1.1.0.

### iOS — File Provider extension (Files app)

- **New `InternxtFileProvider` extension target** (replicated File Provider) wired into the Xcode project, with domain registration/removal on login/logout (PB-5920).
- **Browse**: enumerate Drive folders in Files.app via tholution through `item(for:)`.
- **Download & open**: files are downloaded and decrypted on demand when opened from Files.
- **Upload & folders**: create folders and upload encrypt
- **File management**: rename, move, and trash support, with errors mapped to standard `NSFileProviderError`.                                  - **Live refresh**: working-set change-feed signaling (`Shot/diff) so Files.app reflects changes made elsewhere; RNdrive mutations (uploads, folder creation, rename) now signal the File Provider.
- **Shared credentials**: `SharedAuthKeychain` extended to JSON-encoded payloads with background-read accessibility, forwarding `driveBaseUrl` and theme preference to the extension via the App Group.
- Extension refactored into focused services (enumeration, download, upload, mutation) for testability.

### Cross-platform / React Native

- New `InternxtSignalingModule` and `InternxtAuthCredentialsModule` JS wrappers; auth slice syncs native credentials on login/logout.
- Drive file/folder services signal the native layer afte
- Assorted fixes: encryption version format in `CreateFileEntry`, `openDocument` exception handling, ETag handling in multipart uploads.


CandelR (commented):
Hey @terrerox , have a look at the SonarCloud issues and the CI tests. Please also add a description of what’s included in this release :)

terrerox:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=internxt_drive-mobile&pullRequest=533) **Quality Gate failed**
> Failed conditions  [C Security Rating on New Code](https://sonarcloud.io/dashboard?id=internxt_drive-mobile&pullRequest=533) (required ≥ A)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=internxt_drive-mobile&pullRequest=533)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

1. AndroidManifest.xml (S6361): wants separate read/write permissions on the provider, but Android forces DocumentsProviders to use MANAGE_DOCUMENTS for both. I even tried splitting it into readPermission/writePermission and Sonar still complains, since it's the same permission either way. Not a real issue anyway, only the OS itself can hold that permission.

2. SharedAuthKeychain.swift: asks why the item doesn't require Face ID/passcode. It can't, the File Provider extension reads those credentials in the background with no UI, so there's nowhere to show a prompt. Adding auth there would just break the Files.app integration. The items are still locked to our signed apps only and encrypted until first unlock.

.github/CODEOWNERS
```diff
@@ -4,3 +4,4 @@ src/network/crypto.spec.ts                                   @TamaraFinogina
 src/shareExtension/services/shareEncryptionService.ts        @TamaraFinogina
 src/shareExtension/services/shareUploadService.ts            @TamaraFinogina
 src/network/NetworkFacade.ts                                 @TamaraFinogina
+android/app/src/main/java/com/internxt/cloud/documents/crypto/ @TamaraFinogina
```
CandelR: remove tamara from code owners and merge master please :)

neuffer:
Testing this release from Play (v1.11.0 (1)) against KeePassDX, which is the
first time a third-party app has been able to reach Drive at all. Browsing and
opening work. Saving cannot, and the cause is a single guard rather than
anything structural.

**`openForWrite` only accepts pending uploads.**

```kotlin
// InternxtDocumentsProvider.kt:713
val token = DocumentId.decodeUpload(documentId)
    ?: throw FileNotFoundException("Write only supported on pending uploads: $documentId")
```

`decodeUpload` returns non-null only for the `"u:"` prefix, which
`createPendingFileDocument` mints for a brand-new file. An existing file's id
carries `"d:"`, so `openDocument(<existing file>, "w" | "rw" | "rwt")` throws
before a byte moves. The provider can create a file and stream it up, and after
that its contents are immutable through SAF. Consistent with that,
`FLAG_SUPPORTS_WRITE` does not appear anywhere in the `documents/` package:
`FILE_FLAGS` is `RENAME or DELETE or MOVE`.

For KeePassDX the chain is:

1. it serialises the database to a local cache file, which succeeds;
2. `ContentResolver.getUriOutputStream` calls `openOutputStream(uri, "wt")` and,
   on `FileNotFoundException`, retries `"rwt"`;
3. both contain `w`, so both reach `openForWrite` and both throw;
4. `Database.saveData` wraps it as `DatabaseOutputException`, surfaced as
   "could not save database".

Nothing is corrupted: the refusal happens before any upload starts.

**A second, smaller thing found while tracing that.** In `queryDocument`:

```kotlin
// InternxtDocumentsProvider.kt:150
val row = loadDocumentRow("queryDocument", decoded?.kind, uuid)
cursor.addDocumentRow(row ?: DocumentRowBuilder.folderRow(uuid = uuid, displayName = uuid))
```

`loadDocumentRow` returns null both when credentials are not yet synced and when
the API call throws, and the fallback then reports the document as a **folder
whose display name is its own UUID**. A transient auth or network failure becomes
a plausible-looking wrong answer, and a file is described with `MIME_TYPE_DIR`.
I saw this directly: my database first appeared under its UUID and later, once
the lookup succeeded, under its real name. Returning an empty cursor would let
SAF treat it as not-found instead.

**A possible fix, if it is useful to you:**

https://github.com/neuffer/drive-mobile/tree/fix/saf-write-existing-documents

Six commits on `release-1.9.1` @ `05c8a785`, since this code is not on `master`
or in any tag yet:

1. `FLAG_SUPPORTS_WRITE` on file rows. Note this also corrects
   `DocumentRowBuilderTest.fileRowFields`, which asserts the whole flag word by
   equality and would otherwise go red.
2. `InternxtApiClient.replaceFileContent`, calling `PUT /files/{uuid}`.
   `finishUpload` already returns the new bucket contents id, which is exactly
   what `ReplaceFileDto.fileId` wants, so no new upload machinery is needed.
   `size` and `fileId` travel together because the DTO requires the id when the
   size is positive and forbids it when the size is zero.
3. The existing-file branch in `openForWrite`. A pending upload can stream into
   a pipe; an existing file cannot, because the replacement is only safe once
   the client has finished, so the write lands in a temp file and the upload
   runs from the close callback. That also makes the descriptor seekable, which
   `"rw"` needs. The file keeps its uuid, so document ids and permission grants
   held by other apps survive the write.
4. Two further rounds of review folded in. The ones worth naming, because each
   is a trap in this shape of code rather than a slip:
   - **`UploadForegroundService.start` must be inside the try.** It runs from a
     close callback, when the app may already be backgrounded, and on API 31+ a
     background foreground-service start throws. Outside the try that escapes
     into a scope with no handler, and the temp files and the notification go
     with it.
   - **Whether a mode truncates should be asked of `parseMode`, not restated.**
     I had encoded the javadoc's claim that `"w"` truncates. AOSP's
     `translateModeStringToPosix` sets `O_TRUNC` only when the mode contains
     `'t'`, so `"w"` does not truncate, and a client opening `"w"` to patch
     bytes at an offset would have lost the rest of its file.
   - **Replacements of one file need serialising.** Two saves in flight race,
     and the slower one can land on top of the newer.
5. A trashed file cannot be opened for writing. `replaceFile` refuses anything
   whose status is not `EXISTS`, but `getFileMetadata` calls `findByUuid` with
   no status filter, so a trashed file returns metadata normally and the write
   used to fail only at close, after the client had been told it succeeded. The
   client now parses `status` and refuses the open.

**Please treat this as raw and unverified.** I have no Android toolchain on this
machine, so **none of it has been compiled and the tests I added have never been
run.** It is written from reading your source and it follows the shapes already
in the file, but it should be read as a proposal and a starting point rather
than as something ready to merge. The analysis above is what I am confident in;
the patch is the cheapest way I could make it concrete.

**On why this is a comment and not a pull request.** I am not opening one
deliberately. `drive-desktop-linux` and `drive-server-wip` both refuse pull
requests from forks, and when I raised a server-side issue as
`drive-server-wip#1152` it was closed unmerged with the request that I use
official channels instead. I have taken that as the answer, so the branch is
simply there to pull from or to ignore, whichever suits you. Happy to answer
questions here if any of it is unclear.

CandelR (approved):

PR #552, opened 2026-07-28
[PB-6653] bugfix/Backup status can show as idle/cloud fetching while assets are still uploading

CandelR:
Fixes two backup reliability bugs: an incorrect "idle" status shown during an active upload, and edited photos silently losing their changes on retry.

  ## Summary

  - **`src/store/slices/photos/index.ts`**: `runDiscoveryThunk` no longer overwrites the sync status with `idle` if an upload cycle is already running when it finishes.
  - **`thunks/upload.ts`**: `runUploadThunk` also dispatches `uploading` on its own reentrancy-guard skip path, as a second safeguard against the same race. `buildUploadJobs` now keys the `existingRemoteFileId` decision off `remoteFileId` presence instead of `status`, so a failed edit retry that falls to `status='error'` still replaces the existing Drive file instead of silently uploading (and keeping) the old, pre-edit version.
  - **`network/upload.ts`**: added a diagnostic warning when a file has a zero/invalid size before upload.

### DIFF
--- a/src/network/upload.ts
+++ b/src/network/upload.ts
@@ -27,6 +27,10 @@ export async function uploadFile(
   const stat = await RNFS.stat(filePath);
   const fileSize = stat.size;
 
+  if (fileSize <= 0) {
+    logger.warn(`[Upload] File "${filePath}" has invalid size=${fileSize} (stat: ${JSON.stringify(stat)})`);
+  }
+
   const useMultipart = fileSize > MAX_SIZE_FOR_SINGLE_UPLOAD;
 
   const isAborted = () => params.signal?.aborted === true;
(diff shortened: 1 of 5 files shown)


src/network/upload.ts
```diff
@@ -27,6 +27,10 @@ export async function uploadFile(
   const stat = await RNFS.stat(filePath);
   const fileSize = stat.size;
 
+  if (fileSize <= 0) {
+    logger.warn(`[Upload] File "${filePath}" has invalid size=${fileSize} (stat: ${JSON.stringify(stat)})`);
```
larryrider, line 31: if a file has an invalid size, shouldnt this function stop here or throw an error? instead of only log a warning 🤔

CandelR: Totally, but as we’re currently doing QA of this feature, I wanted to check the QA logs first and add this straight away to see what happens, and then we’ll decide which fix to apply :)

larryrider (approved):

PR #584, opened 2026-09-07
[_] featur/Always encrypt emails

CandelR:
The compose email screen used to decide whether to encrypt by matching recipient domains against a hardcoded list, so a single external recipient sent the whole message in cleartext, including the part addressed to an Internxt user. Now every message is encrypted and the only thing that varies is which key wraps each recipient's copy of the session key, rather than from domain suffixes.

## Summary
  - **`mailCrypto.service.ts`**: `encryptAndSendEmail` rewritten around `resolveWrapKeys`: a recipient the server publishes a key for is wrapped with their own key, a keyless recipient on an external domain is wrapped with `SERVER_PUBLIC_KEY` so `mail-server` can decrypt and deliver over SMTP, and a keyless recipient on an active domain aborts the send instead of receiving a message nobody can open. The sender is always wrapped too, which is what makes the Sent folder readable. 
  - **`recipientKeys.service.ts`**: new cache with a 5 minute TTL and normalized addresses, since the key lookup now sits on the critical path of every send. It fails closed when the server does not answer for a requested address, so a partial
  lookup cannot silently downgrade a message from end-to-end encryption to a server wrap.
  - **`src/services/mail/errors.ts`** and **`src/services/mail/mailDomain.ts`**: domain errors for the five failures the send path can hit, plus the address and domain helpers the send path needs. Domain matching is exact equality against the active domains the server returns, not a suffix check, so neither a lookalike domain nor a subdomain of a served domain counts as internal.
  - **`src/screens/mail/EmailDetailScreen/index.tsx`**: the five `console.warn` calls now go through `logger`.
  - **`src/types/app.ts`** and **`env/`**: new `SERVER_PUBLIC_KEY`.


src/services/mail/errors.ts
xabg2, line 1: Add `Object.setPrototypeOf(this, <ERROR NAME>.prototype)` for each error.

xabg2 (dismissed):

xabg2 (approved):

PR #592, opened 2026-09-11
[PB-6787] feature/Limit and clean up the attachments of every outgoing message

CandelR:
Limit and clean up the attachments of every outgoing message

 ## Summary

  - **`attachmentLimits.ts`**: `MAX_ATTACHMENT_BYTES` as the single definition of the attachment size the server accepts
  - **`mailCrypto.service.ts`**: an attachment over the limit is refused with before anything is encrypted or uploaded. Each upload writes its encrypted copy to a temporary file named after a uuid instead of the attachment name, and deletes it in a `finally` whether the upload succeeds or fails. Attachments are uploaded one after another, since each one is held whole in memory while it is encrypted.
  - **`errors.ts`**: added`AttachmentTooLargeError` and `AttachmentUploadFailedError`
  - **`sendErrors.ts`**: user facing messages for both errors, naming the file and formatting the limit from `MAX_ATTACHMENT_BYTES`
  - **`ComposeEmailScreen`**: the document and photo pickers pass the file size, so the limit applies to attachments picked by the user
  - `MailAttachment` gains an optional `size`, and the new messages are added to `strings.ts` in English and Spanish


src/services/mail/errors.ts
```diff
@@ -58,3 +60,21 @@ export class PrimaryRecipientMissingError extends Error {
     this.name = MailErrorName.PrimaryRecipientMissing;
   }
 }
+
+export class AttachmentTooLargeError extends Error {
```
xabg2, line 64: Remember to use `Object.setPrototypeOf` for the errors, otherwise the `instanceof` check will fail.

xabg2 (approved):

